PR #10, opened 2016-06-30
Partition management

jhorstmann:


src/main/java/org/zalando/fahrschein/EventAlreadyProcessedException.java
```diff
@@ -0,0 +1,15 @@
+package org.zalando.fahrschein;
+
+public class EventAlreadyProcessedException extends Exception {
```
AlexanderYastrebov: Why did you decided to go with checked exceptions?

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
```diff
@@ -1,46 +1,33 @@
...
     }
 
     @Override
     public void onSuccess(String eventName, Cursor cursor) throws IOException {
-        try {
-            template.update(update, eventName, cursor.getPartition(), cursor.getOffset());
-        } catch (DataAccessException e) {
-            throw new IOException(e);
-        }
+        template.queryForObject("SELECT * FROM nakadi_cursor_update(?, ?, ?, ?)", new Object[]{consumerName, eventName, cursor.getPartition(), cursor.getOffset()}, Integer.class);
```
AlexanderYastrebov: SELECT * is bad practice

src/main/java/org/zalando/fahrschein/EventAlreadyProcessedException.java
jhorstmann: > Why did you decided to go with checked exceptions?

So that I can't forget to handle it in `NakadiReader`. It's part of the interface between the `Listener` and `NakadiReader`.

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
jhorstmann: > SELECT * is bad practice

From a table, yes, but this is a sproc that I control.

PR #16, opened 2016-07-01
Fix Jackson Deprecation Warning

lukasniemeier-zalando:
Fixing the blocking issue here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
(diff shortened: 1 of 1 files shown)


jhorstmann:
I actually left the deprecated name on purpose because apart from that constant the library should still be compatible with jackson 2.6, but since we are not testing that version we might as well remove compatibility.

lukasniemeier-zalando:
If you want to signal compatibility with 2.6 then I'd propose to change the dependency to `2.6.7` and stick with the old name (but I'm in for going with `2.7.5` :wink:).

jhorstmann:
Deprecation warning was fixed in #41 by using `SNAKE_CASE`, if someone needs jackson 2.6 compatibility there should be a new issue for it.

lukasniemeier-zalando:
But now the `README` is not in sync anymore. Nooooooooooooooooooooooooooooooooooo :boom:

Issue #17, opened 2016-07-01
Slow Reconnect with HttpComponents


lukasniemeier-zalando:
If a client uses a `HttpComponentsClientHttpRequestFactory` re-connecting on processing failures (i.e.  `IOException`) is very slow. 

_fahrschein_ [tries to close](https://github.com/zalando-incubator/fahrschein/blob/master/src/main/java/org/zalando/fahrschein/NakadiReader.java#L190) the `ClientHttpResponse` in case of a failure. A `HttpComponentsClientHttpResponse` will always try to release the connection back to the connection manager [by consuming the full response body](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpResponse.java#L89) until an _EOF_. This can take an unpredictable amount of time, effectively pausing _fahrschein_.

See [here](https://github.com/lukasniemeier-zalando/fahrschein/blob/apache-blocks-on-close/src/test/java/org/zalando/fahrschein/NakadiReaderIT.java) for an example failing test.

jhorstmann:
At first this looked like an issue in springs `HttpComponentsClientHttpResponse`, but I think this behaviour is built directly into apache httpclient. `EntityUtils` also just closes the input stream:

https://github.com/apache/httpcore/blob/4.4.x/httpcore/src/main/java/org/apache/http/util/EntityUtils.java#L88

The input stream implementation is probably `ChunkedInputStream`, which does the consumption:

https://github.com/apache/httpcore/blob/4.0.x/httpcore/src/main/java/org/apache/http/impl/io/ChunkedInputStream.java#L261

I actually assumed the `JsonParser` to already close the input stream, but it seems I was wrong and it considers the input stream to be managed outside.

lukasniemeier-zalando:
`HttpComponentsClientHttpResponse` does explicitly use the `EntityUtils#consume` before unconditionally releasing the condition via `CloseableHttpResponse#close`.

```java
	@Override
	public void close() {
        // Release underlying connection back to the connection manager
        try {
            try {
                // Attempt to keep connection alive by consuming its remaining content
                EntityUtils.consume(this.httpResponse.getEntity());
            }
			finally {
                this.httpResponse.close();
            }
        }
        catch (IOException ex) {
			// Ignore exception on close...
        }
}
```

[...]

Issue #15, opened 2016-07-01
Simplify building of ClientHttpRequestFactory and NakadiClient itself


lukasniemeier-zalando:
Clients have to provide a `requestFactory` to the `NakadiClient`, which can handle authorization and `problem`-responses.

``` java
final SimpleClientHttpRequestFactory requestFactoryDelegate = new SimpleClientHttpRequestFactory();
final ProblemHandlingClientHttpRequestFactory problemHandlingRequestFactory = new ProblemHandlingClientHttpRequestFactory(requestFactoryDelegate, objectMapper);
final ClientHttpRequestFactory requestFactory = new AuthorizedClientHttpRequestFactory(problemHandlingRequestFactory, () -> "MY_ACCESS_TOKEN");
```

As both is [mandatory](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml) when communicating with _nakadi_ it would be great if _fahrschein_ could offer an easy way to wrap a client's `requestFactory`.

jhorstmann:
I introduced this in the latest commit of PR #18 

https://github.com/zalando-incubator/fahrschein/blob/close-http-components/src/main/java/org/zalando/fahrschein/ClientHttpRequestFactoryHelper.java

I don't like the name though, especially as we discussed having no "helpers" or "utils" so far :)

lukasniemeier-zalando:
We could hide this in a `NakadiClientBuilder` like this:

```java
client = NakadiClient.Builder.for("https://my.nakadi.zalan.do")
    .withBackoff(new ExponentialBackoff())
    .withCursorManager(new PersistentCursorManager())
    .wrappingRequestFactory(new HttpComponentsRequestFactory())
    .build();
```

`ExponentialBackoff`, `inMemoryCursorManager` and `SimpleRequestFactory` can be defaults.

jhorstmann:
One complication is that the `ManagedCursorManager` itself needs a `ClientHttpRequestFactory`, so once you use that you again have to configure everything yourself.

tmuehl:
Why not creating the HttpClient inside `NakadiClient` instead of asking the user for it? We can let the user specify the Parameters (e.g. connection and read timeouts), which then could be also coordinated with the Nakadi parameters `stream_timeout`, `batch_flush_timeout` etc.

PR #18, opened 2016-07-02
Integration test for reconnection issue with apache http client

jhorstmann:


src/main/java/org/zalando/fahrschein/ClientHttpRequestFactoryHelper.java
```diff
@@ -0,0 +1,34 @@
+package org.zalando.fahrschein;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+
+public class ClientHttpRequestFactoryHelper {
```
tmuehl: I don't get that class and wrapping. 😟 
And it's also only used in tests but placed in production code.

jhorstmann: > I don't get that class and wrapping. :worried: Can you tell me, what's the purpose?

The idea was to automatically wrap a request factory so it does authorization, problem handling and work around an issue with apache http client. For problems and authorization there is a better idea in issue #15 (using a builder for the nakadi client), the issue with apache http client can be worked around outside the library as described in #17.

PR #19, opened 2016-07-03
Custom apache request factory

lukasniemeier-zalando:
Referencing #17 and #18.

Replacing the reflection based request factory patch in #18 with a minimal re-implementation of the spring glue code.


jhorstmann:
I like the idea of solving this without reflection, and also that it is more explicit about having different behaviour, compared to silently wrapping. I'd prefer overriding some methods of the original implementation, but it seems that implementations contains some package-private and final classes.

Your implementation might have slightly different behaviour in other places which might confuse users. Adding it to the core library would also add a compile-time dependency to apache http client.

Long-term solution could be to create a bug report and pull request for spring framework to make the behaviour configurable, or at least make the implemetation more extensible.

lukasniemeier-zalando:
> Your implementation might have slightly different behaviour in other places which might confuse users. 

You mean if users would use this in other places than *fahrschein*? 

> Adding it to the core library would also add a compile-time dependency to apache http client.

So would (should?) a reflection based solution. I see no way around providing a work around for HttpComponents in *fahrschein* for now.

lukasniemeier-zalando:
If we do not want to have this in *fahrschein* - should we add a hint to the README?

jhorstmann:
> should we add a hint to the README?

Done in PR #22, I'm going to close this as wontfix since it could be done outside fahrschein, ok?

PR #20, opened 2016-07-03
Allow Custom Mapping Error Handling

### LINKED ISSUE #9: Support passing a custom error handler to NakadiClient / NakadiReader
As a user a NakadiClient
I want to pass a own implementation of a error handler
So that I can decide how to handle different kind of exceptions.

In the current implementation NakadiClient catches all IOExceptions and does retries.
This behaviour is somehow unfortunate, when reading an event that can't be deserialized by the provided ObjectMapper. Example:
```
2016-06-29 13:56:58,604 {WARN} [NakadiEventListener-1] [] [org.zalando.fahrschein.NakadiReader] Got [JsonMappingException] while reading events
com.fasterxml.jackson.databind.JsonMappingException: Failed to deserialize java.time.OffsetDateTime: (java.time.format.DateTimeParseException) Text '2016-06-28T14:41:59' could not be parsed at index 19
```
NakadiClient will then end up in an (in)finit reconnect-retry-loop.

lukasniemeier-zalando:
Fixes #9.

* allow for partial parsing of an event batch
* user method decides whether to continue or abort connection on parse error
* added crude test to show this off


src/main/java/org/zalando/fahrschein/Listener.java
```diff
@@ -4,6 +4,9 @@
 import java.util.List;
 
 public interface Listener<T> {
-    void onEvent(List<T> event) throws IOException, EventAlreadyProcessedException;
+    void onEvent(final List<T> event) throws IOException, EventAlreadyProcessedException;
 
+    default void onError(final IOException e) throws IOException {
+        // no operation
```
jhorstmann: Default should probably be to throw the exceptions instead of silently swallowing it.

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -121,9 +122,20 @@ private Cursor readCursor(JsonParser jsonParser) throws IOException {
...
 
-        while (jsonParser.nextToken() == JsonToken.START_OBJECT) {
-            final T event = objectReader.readValue(jsonParser, eventClass);
-            events.add(event);
+        if (jsonParser.nextToken() == JsonToken.START_OBJECT) {
+            final Iterator<T> eventIterator = objectReader.readValues(jsonParser, eventClass);
+            while (eventIterator.hasNext()) {
+                try {
+                    events.add(eventIterator.next());
+                } catch (final Exception e) {
```
jhorstmann: This could be more explicit by only catching `RuntimeException` or even `RuntimeJsonMappingException`. There is another `onError` method in `CursorManager` so maybe it makes sense for this to only handle mapping exceptions.

jhorstmann: `JsonMappingException` even contains some additional information which might make it possible to handle the error in a sensible way or only ignore certain errors.

On the other hand, if you know where you expect errors, it might be better to just have a more lenient deserializer.

src/main/java/org/zalando/fahrschein/Listener.java
lukasniemeier-zalando: I agree. I will adapt this - I might not get arround writing actual Tests...

src/main/java/org/zalando/fahrschein/NakadiReader.java
lukasniemeier-zalando: Ok we will switch to `JsonMappingException` leaving out `JsonParseException`s, which make sense IMO. I will clarify that users only have to handle mapping errors.

lukasniemeier-zalando:
I am waiting on the master to be ready for this to be rebased, please ping me.

PR #29, opened 2016-07-05
Retry connecting in case of read timeouts during json parser creation

### LINKED ISSUE #27: ReadTimeout during Reconnection is not handled
In `NakadiReader` the `catch` guards only opening the stream. `createParser` will try to read some bytes from the stream, effectively running into a `ReadTimeout` if nothing happens.

```java
                try {
                    LOG.info("Reconnecting after [{}] errors", errorCount);
                    try {
                        response = backoffStrategy.call(errorCount, e, this::openStream);
                    } catch (BackoffException e1) {
                        LOG.warn("Could not reconnect after [{}] errors", errorCount, e1);
                        return;
                    }
                    jsonParser = jsonFactory.createParser(response.getBody());
                } catch (InterruptedException e1) {
                    LOG.warn("Interrupted during reconnection");

                    Thread.currentThread().interrupt();
                    return;
                }
```

jhorstmann:
Fixes #27

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiReader.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiReader.java
@@ -19,6 +19,7 @@ import org.zalando.fahrschein.domain.Batch;
 import org.zalando.fahrschein.domain.Cursor;
 import org.zalando.fahrschein.domain.Subscription;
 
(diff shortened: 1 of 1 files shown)


lukasniemeier-zalando:
:+1:

PR #32, opened 2016-07-06
reducing log levels based on exception

### LINKED ISSUE #31: Reduce Log Spam on Remote Closing the Stream
Currently there is a warning log whenever remote closes the stream.

Currently...

* `NakadiReader` [waits for the next cursor to start](https://github.com/zalando-incubator/fahrschein/blob/master/src/main/java/org/zalando/fahrschein/NakadiReader.java#L215)
* If a `null` value is returned from the stream a generic `IOException` is [thrown](https://github.com/zalando-incubator/fahrschein/blob/master/src/main/java/org/zalando/fahrschein/NakadiReader.java#L279)
* The exception is caught and [logged as a warning](https://github.com/zalando-incubator/fahrschein/blob/master/src/main/java/org/zalando/fahrschein/NakadiReader.java#L239)
* As expected the reconnection logic is triggered

What I'd like to see..

* A specific `StreamClosedException extends IOException` is thrown`
* `IOException`s of this kind are logged at INFO level without the stack trace

lukasniemeier-zalando:
Fixes #31.

Something like this could be a start...

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiReader.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiReader.java
@@ -212,7 +212,7 @@ public class NakadiReader<T> {
 
                 LOG.debug("Waiting for next batch of events for [{}]", eventName);
 
-                expectToken(jsonParser, JsonToken.START_OBJECT);
+                expectBatchStartToken(jsonParser, JsonToken.START_OBJECT);
                 expectToken(jsonParser, JsonToken.FIELD_NAME);
                 expectField(jsonParser, "cursor");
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/fahrschein/StreamClosedException.java
```diff
@@ -0,0 +1,10 @@
+package org.zalando.fahrschein;
+
+import java.io.IOException;
+
+public class StreamClosedException extends IOException {
```
jhorstmann: I was thinking about reusing `java.io.EOFException` for this.

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -265,10 +264,22 @@ public void run(long timeout, TimeUnit timeoutUnit) throws IOException {
...
+        } else {
+            LOG.warn("Got [{}] while reading events", e.getClass().getSimpleName(), e);
+        }
+    }
+
     private void expectField(JsonParser jsonParser, String expectedFieldName) throws IOException {
         final String fieldName = jsonParser.getCurrentName();
         if (fieldName == null) {
-            throw new IOException("Stream was closed or no field at current position");
+            throw new IOException("No field at current position");
```
jhorstmann: I'm no longer sure if this can happen at all, if Jackson reaches eof in the middle of the input it should throw

    JsonParseException: Unexpected end-of-input: ...

lukasniemeier-zalando:
Solved in #33

PR #35, opened 2016-07-07
Issue 34 fix content type detection

tmuehl:
* Recognize `application/problem+json` content type correctly.
* Don't try to deserialize `application/json` into a problem, because it isn't necessarily one.
* Write custom deserializer for `IOProblem` because the default one fails

@jhorstmann


jhorstmann:
:+1:

jhorstmann:
:+1:

PR #39, opened 2016-07-07
Handle oauth problems

### LINKED ISSUE #38: Handle oauth error responses and wrap them in problems
Some error responses from nakadi are not send as `application/problem+json` but as `application/json` in the following format:

    {"error":"invalid_request","error_description":"Access Token not valid"}

This is done by spring-security in `OAuth2ExceptionJackson2Serializer`.

jhorstmann:
Fixes #38


src/main/java/org/zalando/fahrschein/ProblemHandlingClientHttpRequest.java
```diff
@@ -1,76 +1,108 @@
...
+            return new IOProblem(URI.create(type), title, status, detail, instance == null ? null : URI.create(instance));
+        } else {
+            final JsonNode errorNode = rootNode.get("error");
+            final JsonNode descriptionNode = rootNode.get("error_description");
+
+            if (errorNode != null && descriptionNode != null) {
+                final String error = errorNode.asText();
+                final String description = descriptionNode.asText();
+
+                return new IOProblem(DEFAULT_PROBLEM_TYPE, error, status, description);
```
lukasniemeier-zalando: You could define a `oauth-problem` here instead of using the default type

lukasniemeier-zalando:
:+1:

src/main/java/org/zalando/fahrschein/ProblemHandlingClientHttpRequest.java
jhorstmann: > You could define a oauth-problem here instead of using the default type

Do you know of a standard URI that could be used?

lukasniemeier-zalando: No you'd have to come up with one probably :cry:

PR #41, opened 2016-07-08
Redis cursor manager

tmuehl:
@jhorstmann @whiskeysierra 
This is a CursorManager for Redis (i.e. AWS ElastiCache as well).

Some Refactorings that come along with it:
* Rename `PersistentCursorManager` to `JdbcCursorManager` because Redis is also persistent (as well as every other storage that will be supported). What is specific about the former `PersistentCursorManager` is that it is using Jdbc.
* Move `CursorManager` and associated classes into a dedicated package. Since the project is growing that brings imho more overview. Raise your voice if you don't like it.

Maybe it would make sense to not deliver the different Cursor / Partition Managers with the "fahrschein-core" artifact but to create additional artifacts: "fahrschein-jdbc-persistence", "fahrschein-redis-persistence". That would avoid too many heavy unused dependencies. E.g. a user who wants to persist to a SQL database is not interested in Redis dependencies. Or do you think the project is too small so far to legitimate different artifacts?


jhorstmann:
Very nice. I agree about the renaming to `JdbcCursorManager`, but would have kept the interface `CursorManager` in the main package, with subpackages for jdbc and redis.

I thought about splitting up the library before, for now I think it would be simpler to mark the redis dependency as `optional` (same as for `spring-jdbc`).

Is the `org.json` dependency needed? It does not look like it's used at all.

The `commons-lang` dependency also seems to be only used in one place and could be replaced by guava `Splitter` or `String.split`.

lukasniemeier-zalando:
> Maybe it would make sense to not deliver the different Cursor / Partition Managers with the "fahrschein-core" artifact but to create additional artifacts: "fahrschein-jdbc-persistence", "fahrschein-redis-persistence". That would avoid too many heavy unused dependencies. E.g. a user who wants to persist to a SQL database is not interested in Redis dependencies.

:+1:

tmuehl:
@jhorstmann I updated the Pull Request as you requested:

* Moved `CursorManager` to top level package again and have deparate `jdbc` and `redis`packages
* Made the Redis dependencies optional. I wasn't aware of this, thanks for pointing out. 👍 
* Got rid of commons-io and replace it with the Guava pendants. Also wasn't aware of them, since I am not so familiar with Guava. Thanks!
* Got rid of unused org.json dependency. (I had a class with a main method for local testing where I used it and forgot to delete it afterwards.)

jhorstmann:
LGTM

One final question, do you rely on the `equals`/`hashCode` methods now in `Cursor`? I agree it's an obvious candidate for being a value object, but I'm very hesitant to adding `equals` methods without good reasons. For example, since the cursor does not contain the event name, two cursors for different events could now compare equals.

tmuehl:
> One final question, do you rely on the equals/hashCode methods now in Cursor? I agree it's an obvious candidate for being a value object, but I'm very hesitant to adding equals methods without good reasons. For example, since the cursor does not contain the event name, two cursors for different events could now compare equals.

I need them for tests only: https://github.com/zalando-incubator/fahrschein/pull/41/files#diff-3583663db788eb83d30d20285ac6468eR39
```
assertThat("Could not serialize and deserialize cursor " + cursor.toString(),
    actualCursor, IsEqual.equalTo(cursor));
```

Fur sure, I could instead write a custom Hamcrest Matcher. But I would personally consider two cursors as equal if the partition name and offset is equal, since the event type name is not part of a cursor.

src/test/java/org/zalando/fahrschein/redis/RedisCursorManagerIT.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.hamcrest.collection.IsEmptyCollection.empty;
+import static org.hamcrest.collection.IsIterableContainingInAnyOrder.containsInAnyOrder;
+import static org.junit.Assert.assertThat;
+
+public class RedisCursorManagerIT {
+
+    public static final String EVENT_TYPE_NAME = generateUniqueEventType();
+
+    @Test
+    @Ignore("Meant for local testing. You need a running redis cluster on localhost.")
```
jhorstmann: How easy is it to start a local redis instance, for example in a docker container?

I tried to automate this for postgres inside the mvn pre-/post-integration-test phases using a docker-maven-plugin (https://fabric8io.github.io/docker-maven-plugin/), but could not connect from the test to the db. Starting the same docker image manually worked.

jhorstmann:
LGTM

src/test/java/org/zalando/fahrschein/redis/RedisCursorManagerIT.java
tmuehl: I guess it should be fairly easy to start it in a docker, I haven't tried to so far. But I can do, if you want to?

PR #42, opened 2016-07-08
Support usage of PersistentCurosorManger with database schemas.

### LINKED ISSUE #40: Support schema for PersistentCursorManager
```org.zalando.fahrschein.PersistentCursorManager``` executes all sql statements against the global schema. It would be nice to tell the ```PersistentCursorManager``` against which schema the sql statements should be executed.

### PR DESCRIPTION
**Pull request by matusewicz**
Fixes #40

### DIFF
--- a/src/main/java/org/zalando/fahrschein/jdbc/JdbcCursorManager.java
+++ b/src/main/java/org/zalando/fahrschein/jdbc/JdbcCursorManager.java
@@ -10,37 +10,53 @@ import javax.sql.DataSource;
 import java.io.IOException;
 import java.util.Collection;
 
+import static java.lang.String.format;
+
 public class JdbcCursorManager implements CursorManager {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcCursorManager.class);
 
     private final JdbcTemplate template;
     private final String consumerName;
+    private final String schemaPrefix;
 
-    public JdbcCursorManager(JdbcTemplate template, String consumerName) {
+    public JdbcCursorManager(final JdbcTemplate template, final String consumerName, final String schema) {
         this.template = template;
         this.consumerName = consumerName;
+        if (schema == null || schema.isEmpty()) {
+            throw new IllegalArgumentException("schema may not be null or empty.");
+        }
+        this.schemaPrefix = schema + ".";
     }
 
-    public JdbcCursorManager(DataSource dataSource, String consumerName) {
+    public JdbcCursorManager(final JdbcTemplate template, final String consumerName) {
+        this.template = template;
+        this.consumerName = consumerName;
+        this.schemaPrefix = "";
+    }
+
+
+    public JdbcCursorManager(final DataSource dataSource, final String consumerName) {
         this(new JdbcTemplate(dataSource), consumerName);
     }
 
     @Override
-    public void onSuccess(String eventName, Cursor cursor) throws IOException {
-        template.queryForObject("SELECT * FROM nakadi_cursor_update(?, ?, ?, ?)",
-                new Object[]{consumerName, eventName, cursor.getPartition(), cursor.getOffset()},
-                Integer.class);
+    public void onSuccess(final String eventName, final Cursor cursor) throws IOException {
+        final String sql = format("SELECT * FROM %snakadi_cursor_update(?, ?, ?, ?)", schemaPrefix);
+
+        template.queryForObject(sql, new Object[]{consumerName, eventName, cursor.getPartition(), cursor.getOffset()}, Integer.class);
     }
 
     @Override
-    public void onError(String eventName, Cursor cursor, Throwable throwable) {
+    public void onError(final String eventName, final Cursor cursor, final Throwable throwable) {
         LOG.warn("Exception while processing events for [{}] on partition [{}] at offset [{}]", eventName, cursor.getPartition(), cursor.getOffset(), throwable);
     }
 
     @Override
-    public Collection<Cursor> getCursors(String eventName) throws IOException {
-        return template.query("SELECT * FROM nakadi_cursor_find_by_event_name(?, ?)", new Object[]{consumerName, eventName}, (resultSet, i) -> {
+    public Collection<Cursor> getCursors(final String eventName) throws IOException {
+        final String sql = format("SELECT * FROM %snakadi_cursor_find_by_event_name(?, ?)", schemaPrefix);
+
+        return template.query(sql, new Object[]{consumerName, eventName}, (resultSet, i) -> {
             return new Cursor(resultSet.getString(2), resultSet.getString(3));
         });
     }
--- a/src/test/java/org/zalando/fahrschein/jdbc/JdbcCursorManagerIT.java
+++ b/src/test/java/org/zalando/fahrschein/jdbc/JdbcCursorManagerIT.java
@@ -14,7 +14,7 @@ import org.zalando.fahrschein.LocalPostgresConfiguration;
 @ContextConfiguration(classes = LocalPostgresConfiguration.class)
 @TransactionConfiguration(defaultRollback = true)
 @Transactional
-public class PersistentCursorManagerIT extends AbstractCursorManagerTest {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/main/java/org/zalando/fahrschein/PersistentCursorManager.java [hunk 1]
```diff
@@ -8,36 +8,53 @@
...
         this(new JdbcTemplate(dataSource), consumerName);
     }
 
     @Override
-    public void onSuccess(String eventName, Cursor cursor) throws IOException {
-        template.queryForObject("SELECT * FROM nakadi_cursor_update(?, ?, ?, ?)", new Object[]{consumerName, eventName, cursor.getPartition(), cursor.getOffset()}, Integer.class);
+    public void onSuccess(final String eventName, final Cursor cursor) throws IOException {
+        final String sql = schema
+                .map(name -> "SELECT " + name + ".nakadi_cursor_update(?, ?, ?, ?)")
+                .orElse("SELECT nakadi_cursor_update(?, ?, ?, ?)");
```
lukasniemeier-zalando: `format(sql, schema.map(s -> s + ".").orElse(""))`

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
```diff
@@ -8,36 +8,51 @@
...
+
+    public PersistentCursorManager(final DataSource dataSource, final String consumerName) {
         this(new JdbcTemplate(dataSource), consumerName);
     }
 
     @Override
-    public void onSuccess(String eventName, Cursor cursor) throws IOException {
-        template.queryForObject("SELECT * FROM nakadi_cursor_update(?, ?, ?, ?)", new Object[]{consumerName, eventName, cursor.getPartition(), cursor.getOffset()}, Integer.class);
+    public void onSuccess(final String eventName, final Cursor cursor) throws IOException {
+        final String sql = format("SELECT {0}nakadi_cursor_update(?, ?, ?, ?)", schema.map(s -> s + ".").orElse(""));
```
tmuehl: Maybe compute a field `schemaPrefix` in the constructor to avoid the `.map().orElse()` repetition with every call.

jhorstmann: I'd also prefer `String.format` instead of `MessageFormat`

src/main/java/org/zalando/fahrschein/domain/Cursor.java
```diff
@@ -1,6 +1,6 @@
 package org.zalando.fahrschein.domain;
 
-public final class Cursor {
+public class Cursor {
```
jhorstmann: Why no longer `final`? There is no need to mock this class since you can just as easily create an instance.

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
matusewicz: ```SELECT %snakadi_cursor_update``` vs. ```SELECT {0}nakadi_cursor_update```
Do you find this more readable?

src/main/java/org/zalando/fahrschein/domain/Cursor.java
matusewicz: There should be alway a possibility to mock away dependencies imho. Otherwise provide a interface.

tmuehl: Can you give an example, where you want to mock a Cursor? It has only 2 `String` fields.

src/test/java/org/zalando/fahrschein/PersistentCursorManagerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        jdbcTemplate = mock(JdbcTemplate.class);
+    }
+
+    @Test
+    public void shouldExecuteSQLStatementForSuccessCallbackWithoutSchemaPrefix() throws Exception {
+
+        new PersistentCursorManager(jdbcTemplate, CONSUMER).onSuccess(EVENT_NAME, mock(Cursor.class));
```
jhorstmann: These tests are very tied to the implementation and will probably break on any refactoring of the `PersistentCursorManager`.

A better test could be similar to `PersistentCursorManagerIT` but with a modified search path of the database.

I would also be ok with merging this simple change without test.

src/main/java/org/zalando/fahrschein/domain/Cursor.java
matusewicz: it's a general thing about not forcing tests code to create unnecessary instances.
Nevertheless, changed back to final, as it makes no sense to argue here about such a simple class.

tmuehl:
If you use a schema for the `CursorManager`, is it worth to also provide a schema to the `PartitionManager`?

jhorstmann:
> If you use a schema for the CursorManager, is it worth to also provide a schema to the PartitionManager?

Could be useful, but `PartitionManager` is still very much in alpha status, I might change the api soon anyway.

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
jhorstmann: > Do you find this more readable?

Not because of readability, but because message format is mostly used for i18n or l10n, the java doc even says

> MessageFormat provides a means to produce concatenated messages in a language-neutral way. **Use this to construct messages displayed for end users.**

tmuehl:
>> If you use a schema for the CursorManager, is it worth to also provide a schema to the PartitionManager?

> Could be useful, but PartitionManager is still very much in alpha status, I might change the api soon anyway.

Ok, I see.

src/test/java/org/zalando/fahrschein/PersistentCursorManagerTest.java
matusewicz: Thats how unit test work. You drive your implementation by writing unit tests when practicing TDD. I would strongly encourage not to replace unit tests with integration tests (test pyramid).

lukasniemeier-zalando:
:+1:

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
```diff
@@ -9,35 +9,49 @@
...
     private static final Logger LOG = LoggerFactory.getLogger(PersistentCursorManager.class);
 
     private final JdbcTemplate template;
     private final String consumerName;
+    private final String schemaPrefix;
 
-    public PersistentCursorManager(JdbcTemplate template, String consumerName) {
+    public PersistentCursorManager(final JdbcTemplate template, final String schema, final String consumerName) {
         this.template = template;
+        this.schemaPrefix = schema + ".";
```
tmuehl: But only

```
if (schema != null && schema.length() > 0) ...
```

:)

lukasniemeier-zalando:
:+1:

src/main/java/org/zalando/fahrschein/PersistentCursorManager.java
```diff
@@ -9,35 +9,52 @@
...
+    private final String schemaPrefix;
 
-    public PersistentCursorManager(JdbcTemplate template, String consumerName) {
+    public PersistentCursorManager(final JdbcTemplate template, final String consumerName, final String schema) {
         this.template = template;
         this.consumerName = consumerName;
+        if (schema == null || schema.isEmpty()) {
+            throw new IllegalArgumentException("schema may not be null or empty.");
+        }
+        this.schemaPrefix = schema + ".";
```
jhorstmann: But the schema can be empty in case it is on the search path. I think you can mark it as nullable and do

    this.schemaPrefix = nullToEmpty(schema) + ".";

tmuehl: Or
```
this.schemaPrefix = isNullOrEmpty(schema) ? "" : schema + ".";
```

matusewicz: I would argue to use the 2 arguments constructor if schema is empty

jhorstmann:
LGTM

PR #45, opened 2016-07-12
Fix potential connection leaks

### LINKED ISSUE #44: Potential Connection Leak on Problem
If a problem is detected and thrown there is a potential connection leak happening [here](https://github.com/zalando-incubator/fahrschein/blob/0.4.0/src/main/java/org/zalando/fahrschein/NakadiReader.java#L120). The `response` is not closed.

jhorstmann:
Fixes #44


lukasniemeier-zalando:
:+1:

PR #50, opened 2016-07-13
Refactor partition manager api

jhorstmann:


src/main/java/org/zalando/fahrschein/PartitionManager.java
```diff
@@ -1,53 +1,28 @@
...
     /**
      *
-     * @param consumerName Name of the consuming application which might be running on multiple nodes
      * @param eventName Name of the event
-     * @param partition The partition to lock
+     * @param partitions The partitions to lock
      * @param lockedBy Name of one node of the consuming application
      * @param timeout Timeout after that the lock will automatically be released
      * @param timeoutUnit Time unit for the timeout
      * @return Whether locking succeeded, i.e. the lock was not held by anyone or the existing lock expired. The current implementations will also return true if the lock is already held by the same node, but not update the expiry time.
```
lukasniemeier-zalando: Update javadoc?

src/main/java/org/zalando/fahrschein/PartitionManager.java
```diff
@@ -1,53 +1,28 @@
...
-    default List<Partition> lockPartitions(String consumerName, String eventName, List<Partition> partitions, String lockedBy, long timeout, TimeUnit timeoutUnit) {
-        final List<Partition> result = new ArrayList<>(partitions.size());
-
-        for (Partition partition : partitions) {
-            final boolean locked = lockPartition(consumerName, eventName, partition.getPartition(), lockedBy, timeout, timeoutUnit);
-            if (locked) {
-                result.add(partition);
-            }
-        }
+    boolean unlockPartitions(Lock lock);
```
lukasniemeier-zalando: adapt javadoc?

lukasniemeier-zalando:
- [ ] document usage of `PartitionManager`

tmuehl:
Looks good 👍

src/main/java/org/zalando/fahrschein/PartitionManager.java
```diff
@@ -9,17 +9,19 @@
...
     /**
+     * Locks partitions for one event. When running consumers on multiple nodes they should try to lock partitions so that events are only processed by one consumer.
      *
      * @param eventName Name of the event
      * @param partitions The partitions to lock
      * @param lockedBy Name of one node of the consuming application
      * @param timeout Timeout after that the lock will automatically be released
      * @param timeoutUnit Time unit for the timeout
-     * @return Whether locking succeeded, i.e. the lock was not held by anyone or the existing lock expired. The current implementations will also return true if the lock is already held by the same node, but not update the expiry time.
+     * @return A {@code Lock} instance containing the list of locked partitions if locking succeded, {@code Optional.empty()} otherwise
```
tmuehl: If you already started with Javadoc, why not `{@link org.zalando.fahrschein.domain.Lock}`?

(you have a type: "succeeded")

tmuehl:
👍

src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -37,13 +38,19 @@
...
+        this(baseUri, clientHttpRequestFactory, backoffStrategy, objectMapper, cursorManager, NO_METRICS_COLLECTOR);
+
+    }
+    public NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, ObjectMapper objectMapper, CursorManager cursorManager, MetricsCollector metricsCollector) {
         this.baseUri = baseUri;
         this.clientHttpRequestFactory = clientHttpRequestFactory;
         this.objectMapper = objectMapper;
         this.cursorManager = cursorManager;
-        this.nakadiReaderFactory = new NakadiReaderFactory(clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper);
+        this.nakadiReaderFactory = new NakadiReaderFactory(clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper, metricsCollector);
```
tmuehl: I explicitly designed the `MetricsCollector` to be passed into the `listen` method. I know this is ugly because you have so many parameters there.

But I wanted to be able to collect different metrics for different eventtypes.

jhorstmann: > I wanted to be able to collect different metrics for different eventtypes.

Ok, I have to look at this again. Would an alternative be to have one MetricsCollector but change the interface so the event name is passed to all methods?

tmuehl: Sounds good! Another alternative might be to have yet another factory, a `MetricsCollectorFactory`.

src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -80,33 +86,30 @@ public Subscription subscribe(String applicationName, String eventName, String c
...
+    public <T> void listen(String eventName, Class<T> eventType, Listener<T> listener, StreamParameters streamParameters, long lockTimeout, TimeUnit timeoutUnit) throws IOException {
         final String queryString = streamParameters.toQueryString();
         final URI uri = baseUri.resolve(String.format("/event-types/%s/events?%s", eventName, queryString));
 
-        final NakadiReader<T> nakadiReader = nakadiReaderFactory.createReader(uri, eventName, Optional.<Subscription>empty(), eventType, listener, metricsCollector);
+        final NakadiReader<T> nakadiReader = nakadiReaderFactory.createReader(uri, eventName, Optional.<Subscription>empty(), eventType, listener);
+
+        if (lockTimeout > 0) {
+            final Thread currentThread = Thread.currentThread();
+            executor.schedule(currentThread::interrupt, lockTimeout, timeoutUnit);
```
tmuehl: Nice approach!

README.md
```diff
@@ -113,7 +113,30 @@ final ManagedCursorManager cursorManager = new ManagedCursorManager(baseUri, req
...
+```
+
+## Using multiple partitions and multiple consumers
+
+With the `PartitionManager` api it is possible to coordinate between multiple nodes of one application, so that only one node is consuming events from a partition at the same time.
 
+Partitions are locked by one node for a certain time. This requires that every node has an unique name or other identifier. The `listen` method will return after the lock timeout and can try to grab the lock again.
+
+```java
+@Scheduled(fixedDelay = 60*1000L)
```
tmuehl: Why do you delay this method?

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -315,7 +303,11 @@ private void expectField(JsonParser jsonParser, String expectedFieldName) throws
     private void expectToken(JsonParser jsonParser, JsonToken expectedToken) throws IOException {
         final JsonToken token = jsonParser.nextToken();
         if (token == null) {
-            throw new EOFException(Thread.currentThread().isInterrupted() ? "Thread was interrupted" : "Stream was closed");
+            if (Thread.currentThread().isInterrupted()) {
```
tmuehl: Doesn't it need to be

```
if (Thread.currentThread().isInterrupted()) {
    throw new InterruptedIOException("Thread was interrupted");
} else {
    if (token == null) {
        throw new EOFException("Stream was closed");
}
```
?
Otherwise the stream won't stop until you also close the connection.

jhorstmann: When the input stream is interrupted its read method will return -1, which will result in a `null` token. The different exception here is just to get nicer messages in logs, the `IOException` will be caught in the main loop and there is another check if the thread was interrupted.

README.md
jhorstmann: > Why do you delay this method?

Good point, there are now two cases where this method can exit:

 - after the specified timeout
 - when an unchecked exception is thrown

For the 2nd case a slight delay might make sense. Another way to implement it would be with the `Async` annotation and having another loop inside the method.

src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: > Out of curiosity, why don't you use stream_timeout?

I want to have control of the timeout on client side. Using the same value for stream and lock timeout makes sense though.

src/main/java/org/zalando/fahrschein/NakadiReader.java
tmuehl: Ah ok, thanks for the explanation. I though the `JsonInput` would block and hence `currentThread().interrupt()` would only set the interrupted flag without killing any connection.

README.md
tmuehl: Ok, I see, makes sense.

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -277,9 +265,9 @@ void runInternal(long timeout, TimeUnit timeoutUnit) throws IOException, Backoff
                 metricsCollector.markErrorWhileConsuming();
 
                 if (errorCount > 0) {
-                    LOG.warn("Got [{}] while reading events for [{}] after [{}] retries", e.getClass().getSimpleName(), eventName, errorCount, e);
+                    LOG.warn("Got [{}] [{}]while reading events for [{}] after [{}] retries", e.getClass().getSimpleName(), e.getMessage(), eventName, errorCount, e);
```
lukasniemeier-zalando: missing whitespace

jhorstmann:
@lukasniemeier-zalando I simplified the api a bit and added a `SimplePartitionManager` which could be used for an implementation based on global resource locks.

PR #53, opened 2016-07-14
Fix cursor manager updating of offsets

### LINKED ISSUE #51: Unable to connect to expired/empty topics
It looks like `0.4.1` introduced a regression concerning expired and empty topics (low-level API).

### Scenario

On...
```
http GET https://nakadi.com/event-types/my-empty-topic/partitions
[
    {
        "newest_available_offset": "BEGIN",
        "oldest_available_offset": "0",
        "partition": "0"
    }
]
```
... with...
```java
cursorManager.updatePartitions(eventName, client.getPartitions(eventName));
client.listen(...);
```
... leads to...
        
    org.zalando.fahrschein.IOProblem: Problem [http://httpstatus.es/412] with status [412]: [Precondition Failed] [partition 0 is empty]

Similar for an expired topic...
```
[
    {
        "newest_available_offset": "1",
        "oldest_available_offset": "2",
        "partition": "0"
    }
]
```
... we get...

    org.zalando.fahrschein.IOProblem: Problem [http://httpstatus.es/412] with status [412]: [Precondition Failed] [offset 2 for partition 0 is unavailable]

### But Why

Might be in ``CursorManager#updatePartitions`` - it is always using the `oldest_available_offset` if no Cursor is available from the manager. Therefore maybe something like this:

    onSuccess(eventName, new Cursor(partition.getPartition(), fallbackToBegin ? "BEGIN" :  partition.getNewestOffset()));

With `fallbackToBegin` being added to the API.

jhorstmann:
Should fix #51


pom.xml
```diff
@@ -331,7 +331,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>1.10.19</version>
+            <version>2.0.86-beta</version>
```
lukasniemeier-zalando: +1000 - they will release a RC very soon.

src/main/java/org/zalando/fahrschein/CursorManager.java
```diff
@@ -41,12 +45,14 @@ default void fromOldestAvailableOffset(String eventName, List<Partition> partiti
...
     default void updatePartitions(String eventName, List<Partition> partitions) throws IOException {
+
         final Map<String, Cursor> cursorsByPartition = getCursors(eventName).stream().collect(Collectors.toMap(Cursor::getPartition, c -> c));
 
         for (Partition partition : partitions) {
+            final String minAvailableOffset = OFFSET_ORDERING.min(partition.getOldestAvailableOffset(), partition.getNewestAvailableOffset());
             final Cursor cursor = cursorsByPartition.get(partition.getPartition());
-            if (cursor == null || !partition.isAvailable(cursor.getOffset())) {
-                onSuccess(eventName, new Cursor(partition.getPartition(), partition.getOldestAvailableOffset()));
+            if (cursor == null || OFFSET_ORDERING.compare(cursor.getOffset(), minAvailableOffset) < 0) {
```
lukasniemeier-zalando: there might be the use case to go for the `maxAvailableOffset` (different issue then).

src/test/java/org/zalando/fahrschein/CursorManagerTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.mockito.Mockito.doCallRealMethod;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class CursorManagerTest {
+
+    /*
+    private static Object[] parameter(String name, @Nullable String initialOffset, String oldestAvailableOffset, String newestAvailableOffset, @Nullable String expectedOffset) {
```
lukasniemeier-zalando: remove?

lukasniemeier-zalando:
:+1:

src/test/java/org/zalando/fahrschein/CursorManagerTest.java
```diff
@@ -0,0 +1,78 @@
...
+        if (expectedOffset != null) {
+            verify(cursorManager).onSuccess("test", new Cursor("0", expectedOffset));
+        } else {
+            verifyNoMoreInteractions(cursorManager);
+        }
+    }
+
+    @Test
+    public void shouldUpdatePartitionWhenNoCursorAndLastConsumedOffsetNoLongerAvailable() throws IOException {
+        run(null, "10", "20", "10");
```
tmuehl: I don't know the purpose of `updatePartitions()` because it's not used, yet. :) But be aware of that if you store it that way it means you have already consumed the offset `"10"`. So if you connect next time with Nakadi offsets:
```
X- Nakadi-Offsets: {"partition": "A", "offset": "10"}
```
the first event you get is the one with offset `"11"`.

jhorstmann:
@lukasniemeier-zalando Can you have a look again?

src/main/java/org/zalando/fahrschein/CursorManager.java
```diff
@@ -32,21 +38,25 @@ default void fromNewestAvailableOffsets(String eventName, List<Partition> partit
...
+     */
     default void updatePartitions(String eventName, List<Partition> partitions) throws IOException {
+
         final Map<String, Cursor> cursorsByPartition = getCursors(eventName).stream().collect(Collectors.toMap(Cursor::getPartition, c -> c));
 
         for (Partition partition : partitions) {
             final Cursor cursor = cursorsByPartition.get(partition.getPartition());
-            if (cursor == null || !partition.isAvailable(cursor.getOffset())) {
-                onSuccess(eventName, new Cursor(partition.getPartition(), partition.getOldestAvailableOffset()));
+            if (cursor == null || (!"BEGIN".equals(cursor.getOffset()) && OFFSET_ORDERING.compare(cursor.getOffset(), partition.getOldestAvailableOffset()) < 0)) {
```
lukasniemeier-zalando: What about extracting `(!"BEGIN".equals(cursor.getOffset()) && OFFSET_ORDERING.compare(cursor.getOffset(), partition.getOldestAvailableOffset()) < 0)` to a meaningful method?

jhorstmann: > What about extracting ... to a meaningful method?

Then I would have to think about a name and place for the method :)

I prefer to not expose it as a public api, which means it can't be in this interface or the domain model.

lukasniemeier-zalando: Maybe it is time for an abstract base class? :wink: I was asking for a extract-to-method for the exact reason that I could not come up with a descriptive name for what this code does :boom:

jhorstmann: > I could not come up with a descriptive name for what this code does :boom:

That would mean the whole `updateCursors` method needs a better name. Something like `skipUnavailableOffsets` perhaps?

jhorstmann:
I created a separate issue #57 for finding a better name, this was supposed to be a bugfix

lukasniemeier-zalando:
:+1:

jhorstmann:
I will release a 0.4.2 release containing this bugfix later

PR #55, opened 2016-07-15
Add jacoco maven plugin

jhorstmann:


pom.xml
```diff
@@ -184,6 +184,52 @@
...
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>report</id>
+                        <goals>
+                            <goal>report</goal>
+                        </goals>
+                    </execution>
+                    <!--
+                    <execution>
```
tmuehl: Should this stay?

tmuehl:
👍

pom.xml
jhorstmann: > Should this stay?

I wanted to avoid having to look up the syntax once we actually have enough coverage to start enforcing it.

jhorstmann:
> :+1:

Looks like you are not part of the zalando or zalando-incubator org and thus zappr does not let you approve the PR.

lukasniemeier-zalando:
:+1:

tmuehl:
> Looks like you are not part of the zalando or zalando-incubator org and thus zappr does not let you approve the PR.

Hmm, actually I am. However, my membership was private, wondering if this is an issue. I changed it. Thanks for pointing out!

PR #56, opened 2016-07-18
Use and log flow

tmuehl:
Our application stopped consuming from Nakadi from one Moment to another without any visible cause. I wanna be able to check with Nakadi logs, which is easier if I have a flow id.

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiClient.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiClient.java
@@ -13,6 +13,7 @@ import org.springframework.http.client.ClientHttpResponse;
 import org.zalando.fahrschein.domain.Partition;
 import org.zalando.fahrschein.domain.Subscription;
 
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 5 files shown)


jhorstmann:
Flow ids are definitely a feature that we need to support. But maybe it can be done in an easier way. If you are using apache http client you could already use the [tracer library](https://github.com/zalando/tracer) which has an interceptor built in. It should also be possible to do this by wrapping `ClientHttpRequestFactory` again, similar the how authorization is handled.

I also have one question, do you expect to reuse the same flow id for reconnections or should a new one be generated? Or should this be up to the user of fahrschein?

tmuehl:
> Flow ids are definitely a feature that we need to support. But maybe it can be done in an easier way. If you are using apache http client you could already use the tracer library which has an interceptor built in. It should also be possible to do this by wrapping ClientHttpRequestFactory again, similar the how authorization is handled.

True, it can be implemented by another HttpRequestWrapper. I will close this one an provide a new PR with using a wrapper. Didn't know about the Tracer library, I will have a look.

> I also have one question, do you expect to reuse the same flow id for reconnections or should a new one be generated? Or should this be up to the user of fahrschein?

I considered the user to provide it to the NakadiClient, because it's hard for the Client to determine if there exists a FlowId in the logging context already. Furthermore, we log against slf4j which has no NDC only a MDC (maybe I am not aware of a work-around), which makes it even harder.

I think, I would reuse the same FlowId when reconnecting because it's still "the same flow". But in the end it doesn't matter, I just want to be able to easily match my logs with Nakadi's logs. :)

PR #60, opened 2016-07-20
Collect metrics

tmuehl:
@jhorstmann @lukasniemeier-zalando @whiskeysierra 

I want to be able to live monitor the client. E.g. I want to be sure, keep alive events still arrive when my listener doesn't consume any events anymore.

Do you have a better idea to collect these metrics?

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiClient.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiClient.java
@@ -12,6 +12,7 @@ import org.springframework.http.client.ClientHttpRequestFactory;
 import org.springframework.http.client.ClientHttpResponse;
 import org.zalando.fahrschein.domain.Partition;
 import org.zalando.fahrschein.domain.Subscription;
+import org.zalando.fahrschein.metrics.MetricsCollector;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -33,13 +34,22 @@ public class NakadiClient {
     private final BackoffStrategy backoffStrategy;
     private final ObjectMapper objectMapper;
     private final CursorManager cursorManager;
+    private final NakadiReaderFactory nakadiReaderFactory;
+    private final MetricsCollector metricsCollector;
 
     public NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, ObjectMapper objectMapper, CursorManager cursorManager) {
+        this(baseUri, clientHttpRequestFactory, backoffStrategy, objectMapper, cursorManager, null);
+    }
+
+    public NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, ObjectMapper objectMapper, CursorManager cursorManager, final MetricsCollector metricsCollector) {
         this.baseUri = baseUri;
         this.clientHttpRequestFactory = clientHttpRequestFactory;
         this.backoffStrategy = backoffStrategy;
         this.objectMapper = objectMapper;
         this.cursorManager = cursorManager;
+        this.metricsCollector = metricsCollector;
+
+        nakadiReaderFactory = new NakadiReaderFactory(baseUri, clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper, metricsCollector);
     }
 
     public List<Partition> getPartitions(String eventName) throws IOException {
@@ -79,7 +89,7 @@ public class NakadiClient {
         final String queryString = streamParameters.toQueryString();
         final URI uri = baseUri.resolve(String.format("/subscriptions/%s/events?%s", subscription.getId(), queryString));
 
-        final NakadiReader<T> nakadiReader = new NakadiReader<>(uri, clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper, eventName, Optional.of(subscription), eventType, listener);
+        final NakadiReader<T> nakadiReader = nakadiReaderFactory.createReader(eventName, Optional.of(subscription), eventType, listener);
 
         nakadiReader.run(streamParameters.getStreamTimeout().orElse(0), TimeUnit.SECONDS);
     }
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -54,6 +55,8 @@
     private final JsonFactory jsonFactory;
     private final ObjectReader eventReader;
 
+    private MetricsCollector metricsCollector = null;
```
jhorstmann: `NakadiReader` was immutable before, I'd like to keep it that way.

src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -33,13 +35,16 @@
     private final BackoffStrategy backoffStrategy;
     private final ObjectMapper objectMapper;
     private final CursorManager cursorManager;
+    private final NakadiReaderFactory nakadiReaderFactory;
 
     public NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, ObjectMapper objectMapper, CursorManager cursorManager) {
         this.baseUri = baseUri;
         this.clientHttpRequestFactory = clientHttpRequestFactory;
         this.backoffStrategy = backoffStrategy;
```
jhorstmann: `BackoffStrategy` can maybe be moved to the factory, I don't think it's used outside of `NakadiReader`

src/main/java/org/zalando/fahrschein/NakadiReaderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.zalando.fahrschein.domain.Subscription;
+import org.zalando.fahrschein.metrics.MetricsCollector;
+
+import javax.annotation.Nullable;
+import java.net.URI;
+import java.util.Optional;
+
+public class NakadiReaderFactory {
```
jhorstmann: Could be package private or do you see a usecase of accessing this from outside the library?

tmuehl: True, it can be package protected.

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -54,7 +56,13 @@
     private final JsonFactory jsonFactory;
     private final ObjectReader eventReader;
 
+    private final Optional<MetricsCollector> metricsCollector;
+
     public NakadiReader(URI uri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, CursorManager cursorManager, ObjectMapper objectMapper, String eventName, Optional<Subscription> subscription, Class<T> eventClass, Listener<T> listener) {
+        this(uri, clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper, eventName, subscription, eventClass, listener, null);
+    }
+
+    public NakadiReader(URI uri, ClientHttpRequestFactory clientHttpRequestFactory, BackoffStrategy backoffStrategy, CursorManager cursorManager, ObjectMapper objectMapper, String eventName, Optional<Subscription> subscription, Class<T> eventClass, Listener<T> listener, @Nullable final MetricsCollector metricsCollector) {
```
tmuehl: It's getting more and more...

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -240,22 +249,30 @@ void runInternal(long timeout, TimeUnit timeoutUnit) throws IOException, Backoff
...
+                    metricsCollector.ifPresent(mc -> mc.markEventsReceived(events.size()));
 
                     expectToken(jsonParser, JsonToken.END_OBJECT);
 
                     final Batch<T> batch = new Batch<>(cursor, Collections.unmodifiableList(events));
 
                     processBatch(batch);
+                } else {
+                    // it's a keep alive batch
+                    metricsCollector.ifPresent(mc -> mc.markEventsReceived(0));
```
jhorstmann: Should the keep alive message count as "events received"? Also, "events received" does not tell us that those events were processed successfully. I would propose the following meters instead:

 - messages received (instead of batches received), this includes keep-alive messages
 - events processed successfully
 - messages failed
 - maybe messages processed (messages received should then equal messages failed + messages processed)

What do you think?

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -54,7 +56,13 @@
     private final JsonFactory jsonFactory;
     private final ObjectReader eventReader;
 
+    private final Optional<MetricsCollector> metricsCollector;
```
jhorstmann: Instead of the `Optional` we could have an empty implementation of the collector, that way we can get rid of all the `ifPresent` checks.

tmuehl: Yes, I was thinking of that solution as well but eventually decided for that `Optional` approach because it can deal with `null` values. In the end I don't care, I can change it.

src/main/java/org/zalando/fahrschein/NakadiReader.java
tmuehl: I counted them as events received because then there is no lack of data if we only receive keep-alive batches for a certain amount of time. Furthermore, you can see on your dashboard that there are really no events inside the messages and everything works fine.

True, I skipped the "eventsSuccessfullyProcessed" meter because I considered it to be monitored by the listener if it could process events or not. Moreover, we are handing over a batch of events to the listener and therefore do not know how many events of that batch were successfully processed and how many were erroneous. We can only tell about the entire batch (message, respectively) whether it was processed or not. We can have a "messagesSuccessfullyProcessed" as you proposed?

> messages received (instead of batches received), this includes keep-alive messages

"messagesReceived" is also fine for me, I change it.

> events processed successfully

See above

> messages failed

You mean renaming "errorWhileConsuming" to "messagesFailed"? I named it "errorWhileConsuming" because you don't really know at this point whether it was an exception in processing a batch or an error in the connection to Nakadi.

> maybe messages processed (messages received should then equal messages failed + messages processed)

I will add this (see above).

jhorstmann:
:+1:

src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -75,11 +78,15 @@ public Subscription subscribe(String applicationName, String eventName, String c
     }
 
     public <T> void listen(Subscription subscription, Class<T> eventType, Listener<T> listener, StreamParameters streamParameters) throws IOException {
+        listen(subscription, eventType, listener, streamParameters, null);
```
lukasniemeier-zalando: why don't you pass here the `NoMetricsCollector`?

tmuehl: I don't know, I don't have a personal preference here. Would you prefer `NoMetricsCollector`?

jhorstmann:
:+1:

PR #70, opened 2016-08-15
Add `@ JsonCreator` to Partition class construtor to be able to parse it using jackson deserializer

mohamedazouz:
* Add `@ JsonCreator` to Partition class construtor to be able to parse it using jackson deserializer


jhorstmann:
:+1:

PR #72, opened 2016-08-16
Unwrap all IOExceptions from MappingIterator

### LINKED ISSUE #71: IOExceptions during reading of events are wrapped in RuntimeExceptions
Happens for example with an `SSLHandshakeException`. The `MappingIterator` wraps it into `RuntimeException` but we only unwrap a very specific `RuntimeJsonMappingException`.

    java.lang.RuntimeException: SSL peer shut down incorrectly  
    at com.fasterxml.jackson.databind.MappingIterator.next(MappingIterator.java:196)  
    at org.zalando.fahrschein.NakadiReader.readEvent(NakadiReader.java:205)

jhorstmann:
This fixes #71

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiReader.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiReader.java
@@ -193,25 +193,26 @@ public class NakadiReader<T> {
         final Iterator<T> eventIterator = eventReader.readValues(jsonParser, eventClass);
 
         final List<T> events = new ArrayList<>();
-        while (eventIterator.hasNext()) {
-            readEvent(eventIterator, events);
-        }
-        return events;
-    }
-
-    private void readEvent(final Iterator<T> source, final List<T> target) throws IOException {
-        try {
-            target.add(source.next());
-        } catch (RuntimeException e) {
-            final Throwable cause = e.getCause();
-            if (cause instanceof JsonMappingException) {
-                listener.onMappingException((JsonMappingException) cause);
-            } else if (cause instanceof IOException) {
-                throw (IOException)cause;
-            } else {
-                throw e;
+        while (true) {
+            try {
+                // MappingIterator#hasNext can theoretically also throw RuntimeExceptions, that's why we use this strange loop structure
+                if (eventIterator.hasNext()) {
+                    events.add(eventIterator.next());
+                } else {
+                    break;
+                }
+            } catch (RuntimeException e) {
+                final Throwable cause = e.getCause();
+                if (cause instanceof JsonMappingException) {
+                    listener.onMappingException((JsonMappingException) cause);
+                } else if (cause instanceof IOException) {
+                    throw (IOException)cause;
+                } else {
+                    throw e;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -200,13 +199,15 @@ private Cursor readCursor(JsonParser jsonParser) throws IOException {
         return events;
     }
 
-    private void readEvent(final Iterator<T> source, final List<T> target) throws JsonMappingException {
+    private void readEvent(final Iterator<T> source, final List<T> target) throws IOException {
         try {
             target.add(source.next());
```
jhorstmann: I'm wondering if the `hasNext` method is also allowed to throw runtime exceptions.

src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -194,23 +193,26 @@ private Cursor readCursor(JsonParser jsonParser) throws IOException {
...
-            target.add(source.next());
-        } catch (final RuntimeJsonMappingException e) {
-            final Throwable cause = e.getCause();
-            if (cause instanceof JsonMappingException) {
-                listener.onMappingException((JsonMappingException) cause);
-            } else {
-                throw e;
+        while (true) {
+            try {
+                // MappingIterator#hasNext can theoretically also throw RuntimeExceptions, that's why we use this strange loop structure
```
lukasniemeier-zalando: But why?

jhorstmann: See https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.7.6/src/main/java/com/fasterxml/jackson/databind/MappingIterator.java#L180 and https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.7.6/src/main/java/com/fasterxml/jackson/databind/MappingIterator.java#L409

Not sure in what cases it can really happen, but according to the code it's possible.

lukasniemeier-zalando: But can we even recover from such cases (i.e. Jackson unable to proceed to next element in events array)?

tmuehl: @lukasniemeier-zalando This can happen if the connection breaks (or is just dropped by Nakadi). In this case we can easily recover by reconnecting. In the catch block he explicitly looks for an IOException, so I think it's fine.

We encounter this exact case in our Team an currently recover by catching this IOException and reconnect.

tmuehl:
👍

src/main/java/org/zalando/fahrschein/NakadiReader.java
lukasniemeier-zalando: Ah now I see `IOException`s are propagated, only `JsonMappingException` are notified.

lukasniemeier-zalando:
:+1:

PR #80, opened 2016-09-06
IOCallable does not need to wrap exceptions

### LINKED ISSUE #58: Why wrapping exception?
Callable.call throws Exception, why the wrapping here?

```java
    default Callable<T> unchecked() {
        return () -> {
            try {
                return call();
            } catch (IOException e) {
                throw new UncheckedIOException(e);
            }
        };
    }
```

jhorstmann:
Fixes #58


lukasniemeier-zalando:
:+1:

Issue #82, opened 2016-09-16
expectToken should throw IOException instead of IllegalStateException


jhorstmann:
The behaviour should be similar to a JsonParseException, triggering retry and backoff

jhorstmann:
Fixed by #84

Issue #81, opened 2016-09-06
Support api changes in high level api


jhorstmann:
New format is currently being discussed in https://github.com/zalando/nakadi/pull/403

The only place where we currently rely on the format is in `CursorManager.updatePartitions` (https://github.com/zalando-incubator/fahrschein/blob/master/src/main/java/org/zalando/fahrschein/CursorManager.java#L52)

It should be possible to handle both formats at the same time, new format should be compared lexicographically, the old format would have to be padded so both offsets have the same length before comparing.

jhorstmann:
- committing cursors requires `X-Nakadi-StreamId` header
 - commit method is `POST` instead of `PUT`
 - format of cursor for committing might have changed to object

jhorstmann:
See https://github.com/zalando/nakadi/pull/436 for more api changes regarding cursor commit

jhorstmann:
Fixed by #84

PR #84, opened 2016-09-22
Ignore unknown properties and subscription api changes

### LINKED ISSUE #77: Support planned metadata on first and last event batch
See https://github.com/zalando/nakadi/pull/406

Unknown properties in the batch should be ignored, or at least the proposed `metadata` property needs to be recognized.

jhorstmann:
This fixes #77, #81 and #82

### DIFF
--- a/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
+++ b/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
@@ -1,12 +1,12 @@
 package org.zalando.fahrschein;
 
-import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Iterables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpMethod;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.client.ClientHttpRequest;
 import org.springframework.http.client.ClientHttpRequestFactory;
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -233,36 +248,56 @@ void runInternal() throws IOException, BackoffException {
...
+                            break;
+                        }
+                        case "metadata": {
+                            LOG.debug("Skipping metadata in event batch");
+                            jsonParser.nextToken();
+                            jsonParser.skipChildren();
+                            break;
+                        }
+                        default: {
+                            LOG.warn("Unexpected field [{}] in event batch", field);
```
lukasniemeier-zalando: Is this just forwards-compatibility or do we expect this already today?

jhorstmann: For compatibility mostly, also according to our api guidelines unknown fields should be ignored.

src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -24,49 +24,109 @@
...
         try (final ClientHttpResponse response = request.execute()) {
 
-            if (response.getStatusCode().value() == HttpStatus.NO_CONTENT.value()) {
-                LOG.warn("Cursor for event [{}] in partition [{}] with offset [{}] was already committed", eventName, cursor.getPartition(), cursor.getOffset());
-            } else if (response.getStatusCode().is2xxSuccessful()) {
-                LOG.debug("Successfully committed cursor for event [{}] in partition [{}] with offset [{}]", eventName, cursor.getPartition(), cursor.getOffset());
+            final int status = response.getStatusCode().value();
+            if (status == 204) {
+                LOG.debug("Successfully committed cursor for subscription [{}] to event [{}] in partition [{}] with offset [{}]", subscriptionId, eventName, cursor.getPartition(), cursor.getOffset());
+            } else if (status == 200) {
```
lukasniemeier-zalando: no more `HttpStatus.XXX`?

src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -24,49 +24,109 @@
...
+                LOG.warn("Cursor for subscription [{}] to event [{}] in partition [{}] with offset [{}] was already committed", subscriptionId, eventName, cursor.getPartition(), cursor.getOffset());
+            } else {
+                LOG.warn("Unexpected status code [{}] for subscription [{}] to event [{}] in partition [{}] with offset [{}]", status, subscriptionId, eventName, cursor.getPartition(), cursor.getOffset());
```
lukasniemeier-zalando: Can one safely continue in case the cursors is not committed?

jhorstmann: It should still be a succesful response, codes >= 400 are handled by `ProblemHandlingClientHttpRequest`. Not sure what would happen with the 3xx range though.

lukasniemeier-zalando:
:+1:

PR #86, opened 2016-10-14
Allow easy http compression

tmuehl:
Provide a ClientHttpRequestFactory wrapper to provide http compression when talking with Nakadi.

### DIFF
--- a/src/main/java/org/zalando/fahrschein/EnableCompressionClientHttpRequestFactory.java
+++ b/src/main/java/org/zalando/fahrschein/EnableCompressionClientHttpRequestFactory.java
@@ -1,0 +1,23 @@
+package org.zalando.fahrschein;
+
+import org.springframework.http.HttpMethod;
+import org.springframework.http.client.ClientHttpRequest;
+import org.springframework.http.client.ClientHttpRequestFactory;
+
+import java.io.IOException;
+import java.net.URI;
+
+public class EnableCompressionClientHttpRequestFactory implements ClientHttpRequestFactory {
+    private final ClientHttpRequestFactory delegate;
+
+    public EnableCompressionClientHttpRequestFactory(final ClientHttpRequestFactory delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public ClientHttpRequest createRequest(final URI uri, final HttpMethod httpMethod) throws IOException {
+        final ClientHttpRequest request = delegate.createRequest(uri, httpMethod);
+        request.getHeaders().add("Accept-Encoding", "gzip");
+        return request;
+    }
+}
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/fahrschein/EnableCompressionClientHttpRequestFactory.java
```diff
@@ -0,0 +1,23 @@
+package org.zalando.fahrschein;
+
+import org.springframework.http.HttpMethod;
+import org.springframework.http.client.ClientHttpRequest;
+import org.springframework.http.client.ClientHttpRequestFactory;
+
+import java.io.IOException;
+import java.net.URI;
+
+public class EnableCompressionClientHttpRequestFactory implements ClientHttpRequestFactory {
```
lukasniemeier-zalando: What about `CompressingClientHttpRequestFactory` ?

jhorstmann:
Would it make sense to always set this header? Json should always compress well and the overhead of gzip is not very big. Only problem could be if some `ClientHttpRequestFactory` had a buggy implementation.

src/main/java/org/zalando/fahrschein/EnableCompressionClientHttpRequestFactory.java
tmuehl: Sure. :) I rename it.

tmuehl:
@jhorstmann Yes, I think there is no reason to *not* compress. Only for very small messages this might become an overhead. But then you have probably no scaling issues anyway.

You mean to always add the header in `NakadiReader`?

mohamedazouz:
Please check this https://github.com/zalando-incubator/fahrschein/issues/92

jhorstmann:
> What about CompressingClientHttpRequestFactory

Actually this class would be **de**compressing the response :)

jhorstmann:
Could you extend this implementation so it also works for `SimpleClientHttpRequestFactory` or other implementations which do not handle `Content-Encoding` transparently?

A small example in the README would also be great, for example this implementation would have to be wrapped inside `ProblemHandlingClientHttpRequest`, not the other way round.

tmuehl:
>> What about CompressingClientHttpRequestFactory

> Actually this class would be decompressing the response :)

Maybe `RequestNakadiToCompressResponseBodyClientHttpRequestFactory`? :)

tmuehl:
> Could you extend this implementation so it also works for SimpleClientHttpRequestFactory or other implementations which do not handle Content-Encoding transparently?

> A small example in the README would also be great, for example this implementation would have to be wrapped inside ProblemHandlingClientHttpRequest, not the other way round.

Sure, I do.

tmuehl:
This branch is outdated. Anyway we wanted to find a new approach.

PR #87, opened 2016-10-14
Provide additional MetricsCollectors

tmuehl:
- `MultiplexingMetricsCollector`: To be able to register multiple `MetricCollector`s:
- `LastActivityMetricsCollector`: In the past we often wanted to build health checks to detect that Fahrschein doesn't consume any events anymore. The Problem with the Dropwizard metric library is that it uses decaying function such that a metric never reaches exact `0`. With this `MetricCollector` one can build alerts like "_Haven't consumed any event since 10 minutes_" and restart the consumer if necessary (sometimes just reconnect helped if the `ClientHttpRequest` stucked silently).

### DIFF
--- a/src/main/java/org/zalando/fahrschein/metrics/LastActivityMetricsCollector.java
+++ b/src/main/java/org/zalando/fahrschein/metrics/LastActivityMetricsCollector.java
@@ -1,0 +1,70 @@
+package org.zalando.fahrschein.metrics;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+
+import java.util.function.LongSupplier;
+
+import static com.codahale.metrics.MetricRegistry.name;
+import static java.lang.System.currentTimeMillis;
+
+public class LastActivityMetricsCollector implements MetricsCollector {
+
+    private long lastMessageReceived = 0;
+    private long lastMessageSuccessfullyProcessed = 0;
+    private long lastEventReceived = 0;
+    private long lastErrorHappend = 0;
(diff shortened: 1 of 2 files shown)


jhorstmann:
I'm thinking about dropping the `MetricsCollector` interface from the api, and instead add the methods with empty default implementations to `Listener`. That way we could remove one of the many parameters to client and reader and it would be possible to implement metrics completely outside the library.

Do you know if other teams are already using the metrics? Even then, the `DropwizardMetricsCollector` would be easy enough to implement as a user.

tmuehl:
True, the many parameters are something we can do better.
On the other hand, if you add the methods to the `Listener` interface, user would have to implement 5 additional methods instead of just creating a given `MetricCollector`. What about adding it to the `StreamParameters`? Or encapsulating everything in a argument object? Or having an `AbstractListener` which takes an optional `MetricsCollector` as an argument and delegates all metrics methods to it?

I saw it was meanwhile refactored anyway that the `MetricsCollector` is given to the `NakadiClient` and not to the `listen()` method. That way I cannot have different metrics for different streams.

We are using it in every project but I don't know about teams. Time for a Fahrschein-Usergroup. 😊

tmuehl:
Another solution: Have a Builder for either the NakadiClient or for listening. E.g.:

```
listen("my.event-type", EventClass.class, myListener)
        .withParameters(streamParameters)
        .withMetricsCollector(metricsCollector)
        .start()
```

jhorstmann:
:+1:

PR #90, opened 2016-10-17
Get rid of too many parameters of the listen() method

tmuehl:
The parameters that are passed in to the `NakadiClient.listen()` method tend to get more and more. For better handling use a DSL to setup listening. One can use it like this:

```
nakadiClient.prepareListening("zalando-sales-offers", SalesOffer.class, new SalesOffersListener())
        .withStreamParameters(myCustomStreamParameters)
        .withMetricCollector(new DropwizardMetricsCollector(metricRegistry, "salesOffers"))
        .startListening();
```

### DIFF
--- a/src/main/java/org/zalando/fahrschein/NakadiClient.java
+++ b/src/main/java/org/zalando/fahrschein/NakadiClient.java
@@ -2,7 +2,6 @@ package org.zalando.fahrschein;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.Iterables;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpMethod;
@@ -10,9 +9,10 @@ import org.springframework.http.MediaType;
 import org.springframework.http.client.ClientHttpRequest;
 import org.springframework.http.client.ClientHttpRequestFactory;
 import org.springframework.http.client.ClientHttpResponse;
+import org.zalando.fahrschein.PreparedListening.LowLevelApiPreparedListening;
+import org.zalando.fahrschein.PreparedListening.SubscriptionApiPreparedListening;
 import org.zalando.fahrschein.domain.Partition;
 import org.zalando.fahrschein.domain.Subscription;
-import org.zalando.fahrschein.metrics.MetricsCollector;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


jhorstmann:
@tmuehl I'm starting to experiment with adding a builder for `NakadiClient` itself. I remember you had a usecase for using different metrics collectors per stream, could this alternatively be solved by either using multiple `NakadiClient` instances or adding the event name to the `MetricsCollector` interface.

Are there usecases for having different `StreamParameter` for the same client but different events? Otherwise I would also like to move this out of the listen method.

tmuehl:
@jhorstmann Different `NakadiClient` instances for different event-types is also fine. Maybe we introduce a `NakadiServer` or `NakadiHost` or similar class which holds parameters related to the Nakadi instance. Moreover, we could rename `NakadiClient` to `NakadiEventTypeClient`. What do you think?

Yes, I want to monitor streams differently. Consider an application is consuming two different event-types `et1` and `et2` (with lowlevel API). `et1` produces 100 events per second while `et2` produces 1-2 events per second. Monitoring them in one graph isn't very valuable. So far, I personally haven't faced an use-case in which I shared a `MetricCollector` between different event-types.

lukasniemeier-zalando:
+1 on API proposed by @tmuehl and +1 on calling it `Builder` as proposed by @jhorstmann - additionally we also have the use case of monitoring two topics in different ways (on the same nakadi host).

jhorstmann:
My api suggestion would look like this:

    final NakadiClient nakadiClient = NakadiClient.builder(NAKADI_URI)
            .withAccessTokenProvider(new ZignAccessTokenProvider())
            .withCursorManager(cursorManager)
            .build(); // defaults to SimpleClientHttpRequestFactory, since it has no further dependencies

    final List<Partition> partitions = nakadiClient.getPartitions(eventName);
    cursorManager.fromOldestAvailableOffset(eventName, partitions);

    nakadiClient.stream(eventName) // or subscription instead of eventName to use high level api
            .withObjectMapper(objectMapper)
            .withBackoffStrategy(new NoBackoffStrategy())
            .withMetricsCollector(myMetricsCollector)
            .withStreamParameters(new StreamParameters().withStreamLimit(10000))
            .listen(MyEvent.class, listener);

This also moves the `BackoffStrategy` and `ObjectMapper` to the reader since both are not needed for the client itself. What do you think about the `stream`/`listen` naming pair?

I have these changes in a local branch already and will open a PR soon. Still have to adapt the README.

tmuehl:
Closed in favour of #114

PR #93, opened 2016-10-18
Extend DataChangeEvent

mastersimon:
- Add generic type T to DataChangeEvent
- Put DataOperation to an enum

### DIFF
--- a/src/main/java/org/zalando/fahrschein/domain/DataOperation.java
+++ b/src/main/java/org/zalando/fahrschein/domain/DataOperation.java
@@ -1,0 +1,28 @@
+package org.zalando.fahrschein.domain;
+
+/**
+ * The type of operation executed on the entity.
+ * <ul>
+ * <li><strong>C</strong>: Creation</li>
+ * <li><strong>U</strong>: Update</li>
+ * <li><strong>D</strong>: Deletion</li>
+ * <li><strong>S</strong>: Snapshot</li>
+ * </ul>
+ */
+enum DataOperation {
+
+    CREATE("C"),
+    UPDATE("U"),
+    DELETE("D"),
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/fahrschein/domain/DataOperation.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * The type of operation executed on the entity.
+ * <ul>
+ * <li><strong>C</strong>: Creation</li>
+ * <li><strong>U</strong>: Update</li>
+ * <li><strong>D</strong>: Deletion</li>
+ * <li><strong>S</strong>: Snapshot</li>
+ * </ul>
+ */
+enum DataOperation {
```
jhorstmann: Does this need any jackson annotations be be deserialized correctly?

mastersimon: Depends. If you add `JsonProperty` to your getter impl, this should work.
```
    @Override
    @JsonProperty("data_op")
    public DataOperation getOperation() {
        return dataOp;
    }
```

jhorstmann: > @JsonProperty("data_op")

Could this be added to the interface instead?

I was thinking about (de-)serializing the enum itself, as `C` instead of `CREATE` for example.

mastersimon: Sure, I can add the `@JsonProperty` to the interface. Should I do the same for
- getDataType
- getOperation
- getMetadata

I could also add `@JsonCreator` to the Metadata constructor.

jhorstmann:
Changes are included in PR #101, please review

PR #95, opened 2016-10-18
Wrap cursors in object when comitting

### LINKED ISSUE #94: Commit endpoint in subscription api requires cursors wrapped in object

jhorstmann:
Fixes #94


mastersimon:
👍

PR #96, opened 2016-10-18
Use Springs HttpStatus

mastersimon:


src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -120,10 +123,10 @@ public void onSuccess(String eventName, Cursor cursor) throws IOException {
 
         try (final ClientHttpResponse response = request.execute()) {
 
-            final int status = response.getStatusCode().value();
```
mastersimon: Interesting. Than you can close this PR.

PR #99, opened 2016-10-24
Added support for jackson types to enable listening for parametric types

danielrohe:
This pull-request allows listening to parametric types like DataChangeEvent<T> where T is of a special type. Clients can provide the Jackson2 JavaType in the listen method.


jhorstmann:
Can you show me a failing test case for this first? I think this could be fixed by PR #93 which adds a type parameter to the `DataChangeEvent` interface, your event would then be a concrete class.

danielrohe:
Will try to create one, but with this approach I can use a generic class DataChangeEvent<T> and not an interface and tell NakadiReader that the T should be serialized to the parameter-type without extending DataChangeEvent<T>.

danielrohe:
Added a test case that fails because of the parametrized type

jhorstmann:
After experimenting with a generic `DataChangeEvent` interface I don't think this is worth the additional complexity. In PR #101 I added a generic base class which should make implementing this on the user side much easier.

Issue #83, opened 2016-09-21
Closing SimpleHttpClientResponse tries to drain response in spring 4.3.x


jhorstmann:
Fahrschein tries to close the `ClientHttpResponse` in case of any exception in order to retry consuming the stream from a clean state. With current spring versions, this blocks trying to consume remaining data, until the nakadi server decides to close the connection.

Related to issue #17 for `HttpComponentsClientHttpResponse`, the same behaviour was added to `SimpleHttpClientResponse` in https://github.com/spring-projects/spring-framework/commit/b947bfe8e99271b4d88a5a5aba0a8960f76b8efd

jhorstmann:
Related discussion in https://jira.spring.io/browse/SPR-14040 and https://jira.spring.io/browse/SPR-14882

lukasniemeier-zalando:
Any update on this? :wink:

jhorstmann:
Yes, watch this space: https://github.com/zalando-incubator/fahrschein-http

Unfortunately dropping the required spring dependency also means rewriting all tests relying on `MockRestServiceServer`.

jhorstmann:
Fixed in [release 0.10.0](https://github.com/zalando-incubator/fahrschein/releases/tag/0.10.0) by including our own implementation

PR #102, opened 2016-10-27
close input stream before entity to circumvent draining

danielrohe:
This change closes the input stream from the HTTP entity upfront, so that the call to close on the HttpResponseEntity can't drain it.


jhorstmann:
Needs a more general fix, see also issues #17 and #83

Issue #105, opened 2016-11-04
Remove requirement for `setSerializationInclusion(JsonInclude.Include.NON_NULL)`


jhorstmann:

jhorstmann:
Fixed by #106

PR #106, opened 2016-11-08
Remove additional setup of object mapper

coders-kitchen:
The changes introduce on the domain entities used for communicating with Nakadi Jackson annotations in order to remove the necessity of the client to specify the ObjectMapper in a specific way.

It also introduces inner classes on them that provide constants of for the fieldnames used in the JSON representation.


jhorstmann:
A better solution would be to use a separate `ObjectMapper` internally and only use the user provided one for the event content. This would also solve issue #105 about `JsonInclude`, we'd no longer need the hack to support `OffsetDateTime` and we could remove all annotations.

The only tricky part would be deserializing data change events, for which I just added an abstract base class.

coders-kitchen:
So the idea is to have the domain classes agnostic from the used JSON de/serializer. That enables Fahrschein to switch it out more easily if necessary. And Fahrschein could use it's own stuff and does not expose an unnecessary dependency to the using app.
 
I will make some changes and update the pull request.

coders-kitchen:
Regarding changing the constructor: Should I make the change compatible for existing users.
Eg. providing other constructors that does not have the ObjectMapper in place and mark the others deprecated.

Or can it be a breaking change (even if this would be a minor one as someone just have to remove the objectMapper from the constructor call) .

jhorstmann:
Users should still be able to specify the `ObjectMapper` to use for events, that would allow them to register additional modules for example for handling monetary amounts. The api does not need to change for this.

coders-kitchen:
Thanks for clarification

coders-kitchen:
I introduced now a internalized object mapper and removed from nearly all entities the jackson related annotations.

The only exception is the metadata entity to circumvent the problem that the user must configure the object mapper otherwise properly.
I was thinking if it would be possible to use two json parser, the internal one for the metadata and the users one for the rest of the event data. But this is not (easy) to solve.

jhorstmann:
Cool, looks good. I'll add a few comments inline.

src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -38,16 +45,26 @@
...
         this.nakadiReaderFactory = new NakadiReaderFactory(clientHttpRequestFactory, backoffStrategy, cursorManager, objectMapper, metricsCollector);
     }
 
+    private ObjectMapper createObjectMapper() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE);
+        objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        objectMapper.registerModules(new Jdk8Module(), new ParameterNamesModule(), new JavaTimeModule());
+        objectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+        objectMapper.disable(MapperFeature.DEFAULT_VIEW_INCLUSION);
```
jhorstmann: I don't think I understand what `DEFAULT_VIEW_INCLUSION` is doing here exactly. Just reading the javadoc, and since there are no `JsonProperty` annotations on the domain classes I would think that all properties are ignored.

src/main/java/org/zalando/fahrschein/domain/Subscription.java
```diff
@@ -21,24 +19,34 @@
...
 
-    @JsonCreator
-    @Hack("Use String type for createdAt so we do not have to require JavaTimeModule to be registered in the ObjectMapper")
-    public Subscription(@JsonProperty("id") String id, @JsonProperty("owning_application") String owningApplication, @JsonProperty("event_types") Set<String> eventTypes, @JsonProperty("consumer_group") String consumerGroup, @JsonProperty("created_at") String createdAt) {
-        this(id, owningApplication, eventTypes, consumerGroup, createdAt == null ? null : OffsetDateTime.parse(createdAt));
-    }
-
-    public Subscription(String owningApplication, Set<String> eventTypes, String consumerGroup) {
-        this(null, owningApplication, eventTypes, consumerGroup, (String)null);
+    @Hack("Necessary to enable Jackson Wrapper working with multiple constructors but without @JsonCreator annotation")
```
jhorstmann: Would marking the constructor (the one with all parameters) as `JsonCreator` also work? That sounds less like a hack to me.

src/main/java/org/zalando/fahrschein/domain/Subscription.java
```diff
@@ -21,24 +19,34 @@
...
     public Subscription(String id, String owningApplication, Set<String> eventTypes, String consumerGroup, OffsetDateTime createdAt) {
         this.id = id;
         this.owningApplication = owningApplication;
-        this.eventTypes = Collections.unmodifiableSet(new HashSet<>(eventTypes));
+        this.eventTypes = determineEventTypesToBeSet(eventTypes);
         this.consumerGroup = consumerGroup;
         this.createdAt = createdAt;
     }
 
+    @Hack("Necessary to prevent NullPointerException on deserialization")
```
jhorstmann: Not really a hack, I would even inline it directly

    this.eventTypes = unmodifiableSet(eventTypes == null ? emptySet() : new HashSet<>(eventTypes));

src/test/java/org/zalando/fahrschein/NakadiReaderDeserializationTest.java
```diff
@@ -109,8 +109,9 @@ public String getName() {
 
         @JsonCreator
         public CustomerChanged(@JsonProperty("metadata") Metadata metadata, @JsonProperty("data_type") String dataType, @JsonProperty("data_op") DataOperation dataOp, @JsonProperty("data") Customer data) {
-            super(metadata, dataType, dataOp, data);
+            super(metadata, DATA_TYPE, dataOp, data);
```
jhorstmann: `DATA_TYPE` should always equal `dataType` here, but it looks a bit strange to ignore the parameter.

jhorstmann:
What do you think about adding a `SubscriptionRequest` class without the `Nullable` properties from `Subscription`? That would simplify the constructor situation there a bit. It would be a small api change though, but the constructor without all parameters is only supposed to be used internally anyway.

jhorstmann:
@mastersimon @danielrohe can you also review and verify the changes to data change events don't break your usage?

src/main/java/org/zalando/fahrschein/NakadiClient.java
coders-kitchen: As far as I got it it is only in affect, when you use views when serializing objects.
In this case it will exclude all properties that are not explicitly annotated as being part of view.

As we don't use views at all, I would remove it.

src/main/java/org/zalando/fahrschein/domain/Subscription.java
coders-kitchen: perfect, thats way better thanks for the hint

coders-kitchen:
Introducing a subscription request is even better than just using the full constructor.
With this approach we can get rid of the hack constructor and make the internal API easier to read. (As no one would ask, why are these two parameters null :) )

README.md
```diff
@@ -41,7 +41,6 @@ final String eventName = "sales-order-service.order-placed";
 
 // Create an ObjectMapper that understands Nakadi naming conventions and problem responses
 final ObjectMapper objectMapper = new ObjectMapper();
-objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
 objectMapper.registerModule(new ProblemModule());
```
coders-kitchen: The ProblemModule is removed

README.md
```diff
@@ -41,7 +41,6 @@ final String eventName = "sales-order-service.order-placed";
 
 // Create an ObjectMapper that understands Nakadi naming conventions and problem responses
 final ObjectMapper objectMapper = new ObjectMapper();
-objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
```
coders-kitchen: Naming strategy is back

src/test/java/org/zalando/fahrschein/NakadiReaderDeserializationTest.java
coders-kitchen: was a faulty revert of a change. fixed

danielrohe:
Actually from what I see it should work. I would even think further and use the internal object mapper for the event-objects, too. In the event objects I would use an object of type JsonNode to store the variable structure and then have a method on the event that takes an object mapper and a class-type and returns the converted JsonNode as this type using objectMapper.convertValue(payload, type), so that the clients can convert the payload inside the listener and the library doesn't need to care about it.

jhorstmann:
> In the event objects I would use an object of type JsonNode

You are already free to do that (more or less) if you implement `AbstractDataChangeEvent<JsonNode>` or a `Listener<JsonNode>`.

jhorstmann:
I found one small issue with the two `Cursor` constructors, but will merge anyway and create a separate PR with the fix and some tests.

jhorstmann:
:+1:

PR #114, opened 2016-11-18
Simplify configuration of NakadiClient and streaming

### LINKED ISSUE #15: Simplify building of ClientHttpRequestFactory and NakadiClient itself

jhorstmann:
Fixes #15 

Based on api proposal by @tmuehl in PR #90


README.md
```diff
@@ -36,32 +38,27 @@ Fahrschein is available in maven central, so you only have to add the following
...
-
-// Create a listener for our event
+// Create a Listener for our event
 final Listener<SalesOrderPlaced> listener = events -> {
     for (SalesOrderPlaced salesOrderPlaced : events) {
         LOG.info("Received sales order [{}]", salesOrderPlaced.getSalesOrder().getOrderNumber());
     }
 };
 
+// Configure client, defaults to using the high level api with ManagedCursorManger and SimpleClientHttpRequestFactory
```
lukasniemeier-zalando: What is the default for the `AccessTokenProvider`?

src/main/java/org/zalando/fahrschein/NakadiClientBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private NakadiClientBuilder(URI baseUri, AccessTokenProvider accessTokenProvider, ClientHttpRequestFactory clientHttpRequestFactory, CursorManager cursorManager) {
+        this.baseUri = baseUri;
+        this.accessTokenProvider = accessTokenProvider;
+        this.clientHttpRequestFactory = clientHttpRequestFactory;
+        this.cursorManager = cursorManager;
+    }
+
+    public NakadiClientBuilder withAccessTokenProvider(AccessTokenProvider accessTokenProvider) {
+        return new NakadiClientBuilder(baseUri, accessTokenProvider, clientHttpRequestFactory, cursorManager);
```
lukasniemeier-zalando: Should the builder be stateful or not? Looks like a mixture here. I'd expect a builder to be stateful, i.e. 
```
this.accessTokenProvider = accessTokenProvider
return this
```

jhorstmann: The builder is supposed to be immutable, the code here should be similar to the one that lombok `Wither` would generate. The fields could be final though.

lukasniemeier-zalando: If you want to go the `Wither` route why have a builder in the first place? You can offer a default constructor/builder method on `NakadiClient` itself and `wither` methods. I still like a stateful `Builder`variant the most as it renders constructing of a client explicit (the issue we wanted to solve).

jhorstmann: With a builder there is one defined place where defaults could be set. Do you think this pattern is too unexpected?

Same pattern was already used before in `StreamParameters`, only difference is that the name does not end in `Builder`.

tmuehl:
👍

PR #122, opened 2016-12-07
Refactor into modules

### LINKED ISSUE #118: Split into modules
- core
 - jdbc
 - redis
 - metrics-dropwizard

jhorstmann:
Fixes #118


tmuehl:
@jhorstmann What do you think about dependency management to have consistent dependency versions over the different modules?

jhorstmann:
> @jhorstmann What do you think about dependency management to have consistent dependency versions over the different modules?

Thinking about it, at the moment it would be duplicating a lot of lines, but it's probably cleaner than defining the properties for version numbers.

pom.xml
```diff
@@ -46,8 +47,23 @@
...
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
+        <version.jackson>2.7.4</version.jackson>
+        <version.spring>4.1.6.RELEASE</version.spring>
+        <version.slf4j>1.7.13</version.slf4j>
+        <version.log4j>2.4.1</version.log4j>
     </properties>
 
+    <modules>
+        <module>fahrschein</module>
```
jhorstmann: I prefer keeping the name of the main artifact the same as in previous versions, it should make updating less confusing.

Issue #124, opened 2016-12-09
No retry if opening the the event stream fails first time


danielrohe:
The first call to openJsonInput in NakadiReader#runInternal is not in the retry-loop. So if this fails there won't be a retry.

tmuehl:
@jhorstmann Can we distinguish between client and nakadi side errors? Then I'd say retrying in case it's not a `4xx` is fine.

lukasniemeier-zalando:
We can do something similar like riptide does, see https://github.com/zalando/riptide/blob/master/riptide-exceptions/src/main/java/org/zalando/riptide/exceptions/ExceptionClassifier.java#L38

jhorstmann:
> We can do something similar like riptide does

That list of exceptions seems really arbitrary.

> Can we distinguish between client and nakadi side errors?

There are two cases that we might want to handle differently.

 - The connection to nakadi itself could not be established on the first try
 - The connection could be established, but the first batch of events does not arrive before the read timeout. The status line and headers probably did arrive in time.

I don't know how the second case could happen, except for a bug in nakadi itself. Retrying the second case would make sense, assuming it only happens if we would really receive an event batch. Retrying in case of authorization or scope errors would not make sense, but it is very unlikey we would get a read timeout for that.

Disabling encoding autodetection as proposed in in #37 and #123 would solve the second case.

@danielrohe Do you agree with the above scenarios?

danielrohe:
The first case is the common connect timeout where nakadi itself is not reachable and the second is the read timeout case where the response is not returned in time.

Retrying the cases depends on whether the fahrschein client functionality includes retrying until the user stops it. It should be clearly stated whether the fahrschein reader handles the retry on error or whether users of the library have to handle this.

jhorstmann:
> It should be clearly stated whether the fahrschein reader handles the retry on error or whether users of the library have to handle this.

So this is more a documentation issue at the moment?

danielrohe:
Yes, if you state that fahrschein reader should handle this, then both cases should be covered, if not then I would expect the fahrschein reader to stop on any network or serialization issue and let the users of the fahrschein reader handle this problem

jhorstmann:
Fixed by #129

PR #129, opened 2016-12-19
Lazily construct JsonParser

### LINKED ISSUE #123: Creation of JsonParser blocks at determining the encoding
The creation of the JsonParser in NakadiReader:145 blocks if the input stream has no char in Jackson2 while trying to determine the encoding.

A solution to circumvent this problem is to determine the charset in the NakadiReader itself and then hand over an InputStreamReader to the createParser method

```java
Charset charset = StandardCharsets.UTF_8;
MediaType contentType = response.getHeaders().getContentType();
if (contentType != null) {
    Charset contentCharset = contentType.getCharSet();
    if (contentCharset != null) {
        charset = contentCharset;
    }
}
final JsonParser jsonParser = jsonFactory.createParser(new InputStreamReader(response.getBody(), charset)).disable(JsonParser.Feature.AUTO_CLOSE_SOURCE);
```

jhorstmann:
so that read timeouts while determining the encoding of the event batch lead to automatic retry.

@danielrohe can you check if this fixes #123 and #124 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,15 +137,24 @@ public void readSalesOrderPlacedEvents() throws IOException {
 }
 ```
 
+## Exception handling
+
+Exception handling while streaming events follows some simple rules
+
+ - `IOException` and its subclasses are treated as temporary failures and will be retried as specified by the `BackoffStrategy`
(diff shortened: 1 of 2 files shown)


jhorstmann:
:+1:

PR #130, opened 2016-12-21
Extend MetricsCollector to be informed about occurred_at timestamps.

i8r:
Addressing https://github.com/zalando-incubator/fahrschein/issues/126
* changing method markEventsReceived of MetricsCollector
* NakadiReader inspects events and checks whether metadata.occurred_at timestamp is present

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/MetricsCollector.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/MetricsCollector.java
@@ -1,9 +1,12 @@
 package org.zalando.fahrschein;
 
+import java.time.OffsetDateTime;
+import java.util.Optional;
+
 public interface MetricsCollector {
     void markMessageReceived();
 
-    void markEventsReceived(int size);
+    void markEventsReceived(int size, Optional<OffsetDateTime> oldestOccurredAt, Optional<OffsetDateTime> latestOccurredAt);
 
     void markErrorWhileConsuming();
 
--- a/fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderTest.java
+++ b/fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderTest.java
@@ -1,6 +1,9 @@
 package org.zalando.fahrschein;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.hamcrest.Matchers;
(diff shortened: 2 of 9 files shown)


i8r:
It's still in progress but feedback is very welcome.

fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderTest.java
```diff
@@ -536,7 +543,66 @@ public void shouldSendCursorsForLockedPartitions() throws IOException {
 
         nakadiReader.run();
 
+        // todo: This assertion is never reached.
         assertEquals(singletonList("[{\"partition\":\"0\",\"offset\":\"0\"},{\"partition\":\"1\",\"offset\":\"10\"}"), headers.get("X-Nakadi-Cursors"));
     }
 
+    @Test
+    public void shouldCallMetricsCollectorWithoutMetadataBeingPresent() throws Exception {
+        final ClientHttpResponse response = mock(ClientHttpResponse.class);
```
i8r: I was thinking of extracting some of the common setup that gets repeated in almost every test to a setup method. How do you feel about it?

i8r:
Addressing https://github.com/zalando-incubator/fahrschein/issues/126

fahrschein/src/main/java/org/zalando/fahrschein/MetricsCollector.java
```diff
@@ -1,9 +1,12 @@
 package org.zalando.fahrschein;
 
+import java.time.OffsetDateTime;
+import java.util.Optional;
+
 public interface MetricsCollector {
     void markMessageReceived();
 
-    void markEventsReceived(int size);
+    void markEventsReceived(int size, Optional<OffsetDateTime> oldestOccurredAt, Optional<OffsetDateTime> latestOccurredAt);
```
i8r: Should I keep the changes to this interface backwards compatible?

i8r:
@jhorstmann I consider this ready for review.

jhorstmann:
Sorry for not reviewing this PR earlier.

I'm a bit hesitant to merge it, as the change introduces some processing and memory overhead for all users of the library, even if they are not interested in the additional metrics.

Would it be possible to collect these metrics inside the `Listener` (or maybe a new `ListenerMetricsAdapter`) instead?

i8r:
The `Listener` only gets mapped object. If I want to do it in the `Listener`, I have to require some interface. Do you have an idea how to reduce the overhead? I couldn't find an alternative to `jsonParser.readValueAsTree()`.

jhorstmann:
You could require an implementation of the existing [Event interface](https://github.com/zalando-incubator/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/domain/Event.java) in your listener.

If the metadata property was one level higher it would be possible to extract it similar to the cursor data, the way it is now I can't think of a version without overhead.

PR #132, opened 2017-01-14
Fix StreamParameters query string

tmuehl:

The query string is broken right now, because the `&` is missing between the individual parameters.


jhorstmann:
:+1:

PR #136, opened 2017-02-14
Support reading from the beginning in the subscription api

### LINKED ISSUE #128: Support reading from the beginning of a subscription
`SubscriptionRequest` should contain the `read_from` property and it should be possible to specify the value in an overload of `NakadiClient.subscribe`.

See https://github.com/zalando/nakadi/blob/master/api/nakadi-event-bus-api.yaml#L1637

jhorstmann:
Fixes #128


tmuehl:
👍

jhorstmann:
:+1:

PR #140, opened 2017-02-21
Support publishing events

### LINKED ISSUE #48: Support posting events to nakadi
Or should this library concentrate only on the consuming part?

jhorstmann:
Fixes #48

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
@@ -9,6 +9,7 @@ import org.springframework.http.MediaType;
 import org.springframework.http.client.ClientHttpRequest;
 import org.springframework.http.client.ClientHttpRequestFactory;
 import org.springframework.http.client.ClientHttpResponse;
+import org.zalando.fahrschein.domain.BatchItemResponse;
 import org.zalando.fahrschein.domain.Partition;
 import org.zalando.fahrschein.domain.Subscription;
 import org.zalando.fahrschein.domain.SubscriptionRequest;
@@ -17,6 +18,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -31,15 +33,17 @@ public class NakadiClient {
     private final URI baseUri;
     private final ClientHttpRequestFactory clientHttpRequestFactory;
     private final ObjectMapper internalObjectMapper;
+    private final ObjectMapper objectMapper;
     private final CursorManager cursorManager;
 
     public static NakadiClientBuilder builder(URI baseUri) {
         return new NakadiClientBuilder(baseUri);
     }
 
-    NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, CursorManager cursorManager) {
+    NakadiClient(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, ObjectMapper objectMapper, CursorManager cursorManager) {
         this.baseUri = baseUri;
         this.clientHttpRequestFactory = clientHttpRequestFactory;
+        this.objectMapper = objectMapper;
         this.internalObjectMapper = DefaultObjectMapper.INSTANCE;
         this.cursorManager = cursorManager;
     }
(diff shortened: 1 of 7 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -54,6 +58,35 @@ public static NakadiClientBuilder builder(URI baseUri) {
...
+        final ClientHttpRequest request = clientHttpRequestFactory.createRequest(uri, HttpMethod.POST);
+
+        request.getHeaders().setContentType(MediaType.APPLICATION_JSON);
+
+        try (final OutputStream body = request.getBody()) {
+            objectMapper.writeValue(body, events);
+        }
+
+        try (final ClientHttpResponse response = request.execute()) {
+            if ((response.getRawStatusCode() == 207 || response.getRawStatusCode() == 422)) {
```
tmuehl: What about a `401`? I know it's not the responsibility of this part of the code to handle authentication, nevertheless this is a valid Nakadi response.

Moreover, what do you think about a catch all for all other response types != 200? I admit, the Nakadi API only defines `200`, `207`, `401` and `422`, but as a client you never know. :)

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -54,6 +58,35 @@ public static NakadiClientBuilder builder(URI baseUri) {
...
+                try (final InputStream is = response.getBody()) {
+                    final BatchItemResponse[] responses = internalObjectMapper.readValue(is, BatchItemResponse[].class);
+                    final List<BatchItemResponse> failed = new ArrayList<>(responses.length);
+                    for (BatchItemResponse batchItemResponse : responses) {
+                        if (!BatchItemResponse.PublishingStatus.SUBMITTED.equals(batchItemResponse.getPublishingStatus())) {
+                            failed.add(batchItemResponse);
+                        }
+                    }
+                    if (!failed.isEmpty()) {
+                        // TODO: attach corresponding events?
```
tmuehl: Yes, sounds good! Any counterarguments to not attach corresponding events?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClientBuilder.java
```diff
@@ -15,39 +16,48 @@
...
     public NakadiClientBuilder withCursorManager(CursorManager cursorManager) {
-        return new NakadiClientBuilder(baseUri, accessTokenProvider, clientHttpRequestFactory, cursorManager);
+        return new NakadiClientBuilder(baseUri, objectMapper, accessTokenProvider, clientHttpRequestFactory, cursorManager);
     }
 
     private ClientHttpRequestFactory defaultClientHttpRequestFactory() {
         final SimpleClientHttpRequestFactory requestFactoryDelegate = new SimpleClientHttpRequestFactory();
         requestFactoryDelegate.setConnectTimeout(DEFAULT_CONNECT_TIMEOUT);
         requestFactoryDelegate.setReadTimeout(DEFAULT_READ_TIMEOUT);
+        requestFactoryDelegate.setBufferRequestBody(true);
```
tmuehl: Why do you avoid streaming?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -54,6 +58,35 @@ public static NakadiClientBuilder builder(URI baseUri) {
...
     }
 
+    public <T> void publish(String eventName, List<T> events) throws IOException {
+        final URI uri = baseUri.resolve(String.format("/event-types/%s/events", eventName));
+        final ClientHttpRequest request = clientHttpRequestFactory.createRequest(uri, HttpMethod.POST);
+
+        request.getHeaders().setContentType(MediaType.APPLICATION_JSON);
+
+        try (final OutputStream body = request.getBody()) {
+            objectMapper.writeValue(body, events);
```
tmuehl: What about setting the `Content-length` header?
Or do you know if it is set automatically by the `ClientHttpRequestFactory`?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: I'd have to check if nakadi returns a specific content type for the `BatchItemResponse`, then we could skip the status code check. Other error status codes should already have been handled by `ProblemHandlingClientHttpRequestFactory`.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClientBuilder.java
jhorstmann: I'll remove this again, it was a left-over when trying to debug some issue.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: Depends on the settings of the `ClientHttpRequestFactory`, it will either use chunked encoding or buffer into a byte array and then send it including the `Content-Type` header.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: Currently I don't require the events to implement any interface, so there is also no standard way to extract the `eid` for correlation. Could either require an interface or do an instanceof check,

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
tmuehl: Since the `ClientHttpRequestFactory` can be configured by the user, there is a chance that a `ProblemHandlingClientHttpRequestFactory` isn't in place. Would that let the publishing fail silently in case of a response != `207` and `422`?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
tmuehl: Hmm, I see.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
tmuehl: Can't we avoid chunked encoding since we know the length in advance and use streaming?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: Yes, that's the big downside of handling status codes problems inside that wrapper, it already caused me some hard to debug issues. The alternative would be to do these status checks in every method doing a request, maybe using a static helper. I'll think about that.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: But we won't know the length before we serialize all events, or am I misunderstanding the issue?

tmuehl: I was thinking about something like:

```
final ByteArrayOutputStream body = new ByteArrayOutputStream();
objectMapper.writeValue(body, events);
request.getHeaders().setContentLength(body.size());
request.getBody().write(body.toByteArray());
```

tmuehl:
👍

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
jhorstmann: I think this is more or less what the ´ClientHttpRequestFactory` is doing by default.

tmuehl:
👍

jhorstmann:
:+1:

Issue #147, opened 2017-03-08
Support subscription with initial cursors


jhorstmann:
See nakadi release notes for 2017W09 and https://github.com/zalando/nakadi/blob/6b2e65b21dd46b414601f2ef14ebf832cd8b3d5d/api/nakadi-event-bus-api.yaml#L1785

AdnanKukuljac:
@jhorstmann Maybe you can close this one?

jhorstmann:
Fixed by #150

PR #150, opened 2017-03-10
Issue 147/spport subscription with inital cursor

AdnanKukuljac:
#147


fahrschein/src/main/java/org/zalando/fahrschein/domain/SubscriptionRequest.java
```diff
@@ -32,16 +32,18 @@ public String toString() {
...
     private final Position readFrom;
+    private final Set<Cursor> initialCursors;
 
-    public SubscriptionRequest(String owningApplication, Set<String> eventTypes, String consumerGroup, Position readFrom) {
+    public SubscriptionRequest(String owningApplication, Set<String> eventTypes, String consumerGroup, Position readFrom, Set<Cursor> initialCursors) {
         this.owningApplication = owningApplication;
         this.eventTypes = unmodifiableSet(eventTypes == null ? emptySet() : new HashSet<>(eventTypes));
         this.consumerGroup = consumerGroup;
         this.readFrom = readFrom;
+        this.initialCursors = initialCursors;
```
jhorstmann: `SubscriptionRequest` is supposed to be immutable, so it would be better to create an unmodifiable copy. I'd also prefer a `List` instead of a `Set`, `Cursor` has a `hashCode` implementation, but I consider that a mistake as it's only used in tests.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -139,8 +134,9 @@ public Subscription subscribe(String applicationName, Set<String> eventNames, St
      * @return
      * @throws IOException
      */
-    public Subscription subscribe(String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom) throws IOException {
-        final SubscriptionRequest subscription = new SubscriptionRequest(applicationName, eventNames, consumerGroup, readFrom);
+    public Subscription subscribe(String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, Set<Cursor> initialCursors) throws IOException {
```
jhorstmann: I'm wondering if a `Map<String, List<Cursor>>` with event name as key makes more sense in the api, to avoid repeating the event name for each partition. The initial cursors are also only valid if `Position` is indeed `cursors`, so maybe there should be some validation in place.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -134,7 +135,11 @@ public Subscription subscribe(String applicationName, Set<String> eventNames, St
      * @return
      * @throws IOException
      */
-    public Subscription subscribe(String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, Set<Cursor> initialCursors) throws IOException {
+    public Subscription subscribe(String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, List<Cursor> initialCursors) throws IOException {
+
+        if(readFrom.equals(SubscriptionRequest.Position.CURSORS)){
```
jhorstmann: Enum values should in general be compared using `==` instead of `equals`, that way the compiler can warn you if accidentally try to compare enums of different types.

fahrschein/src/main/java/org/zalando/fahrschein/domain/SubscriptionRequest.java
```diff
@@ -32,16 +30,18 @@ public String toString() {
...
     private final Position readFrom;
+    private final List<Cursor> initialCursors;
 
-    public SubscriptionRequest(String owningApplication, Set<String> eventTypes, String consumerGroup, Position readFrom) {
+    public SubscriptionRequest(String owningApplication, Set<String> eventTypes, String consumerGroup, Position readFrom, List<Cursor> initialCursors) {
         this.owningApplication = owningApplication;
         this.eventTypes = unmodifiableSet(eventTypes == null ? emptySet() : new HashSet<>(eventTypes));
         this.consumerGroup = consumerGroup;
         this.readFrom = readFrom;
+        this.initialCursors = unmodifiableList((initialCursors == null) ? emptyList() : initialCursors);
```
jhorstmann: This only creates an unmodifiable wrapper, the underlying `initialCursors` parameter could still be modified. Better to create a copy and wrap that in an unmodifiable list.

jhorstmann:
:+1:

PR #154, opened 2017-03-18
Always close connection when end consumption

tmuehl:
In case of another exception than a IOException, the connection wasn't closed and potentially left stalled.

@jhorstmann @AdnanKukuljac

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
@@ -107,7 +107,7 @@ class NakadiReader<T> implements IORunnable {
         @Override
         public void close() {
             try {
-                if (jsonParser != null) {
+                if (jsonParser != null && !jsonParser.isClosed()) {
                     try {
                         LOG.trace("Trying to close json parser");
                         jsonParser.close();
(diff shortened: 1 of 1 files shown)


jhorstmann:
@tmuehl Do you have the stacktrace from where this exception originated? I'm trying to reproduce this in a unit test.

tmuehl:
@jhorstmann If a `RuntimeException` is thrown in the listener, this message is propagated up without closing the http connection. Therefore, you might end up with stalled connections.

```
public class ConnectionClosingTest {
    
    @Test
    public void nakadiReaderClosesConnection() throws Exception {
        final PoolingHttpClientConnectionManager connManager = new PoolingHttpClientConnectionManager();
        connManager.setMaxTotal(3);
        connManager.setDefaultMaxPerRoute(3);

        final CloseableHttpClient httpClient = HttpClientBuilder.create()
                .setConnectionManager(connManager)
                .useSystemProperties().build();

        final HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(httpClient);

        final NakadiClient nakadiClient = new NakadiClientBuilder(new URI("https://nakadi.example.com"))
                .withClientHttpRequestFactory(requestFactory).build();
        
        while (true) {
            nakadiClient.stream("foobar").listen(Foobar.class, events -> {
                throw new RuntimeException("Intentionally thrown.");
            });
        }
    }
    
    private static class Foobar {
        private int id;
        private String text;
    }
}
```

After reconnecting 3 times, you can run out of connections.
Do you have a better proposal?

jhorstmann:
Got it, I had an error in my test setup, but I'll try to fix this in a slightly different way. Thanks a lot for the report!

tmuehl:
@jhorstmann Great, thanks for fixing it! Closed this in favour of #155

PR #161, opened 2017-04-07
Close subscription on request

coders-kitchen:
This change introduces a reader manager that allows the client to stop the nakadireader without exposing it to the client

This change will remove the problem that a user has either to call Thread.currentThread.interrupt() or throw an RuntimeException from the listener implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,9 +206,13 @@ Subscription subscriptions = nakadiClient.subscription(applicationName, eventNam
 
 // Start streaming, the listen call will block and automatically reconnect on IOException
 new Thread(() -> {
-nakadiClient.stream(subscription)
-        .withReaderManager(readerManager)
-        .listen(SalesOrderPlaced.class, listener);  
+  try {
+    nakadiClient.stream(subscription)
+                .withReaderManager(readerManager)
+                .listen(SalesOrderPlaced.class, listener);
+  } catch (IOException e) {
+    e.printStackTrace();
+  }
 }).run();
 
 // let it run for 1 minute
@@ -217,17 +221,17 @@ Thread.sleep(60*1000);
 // and stop the reading
 readerManager.discontinueReading();
 
-// let's wait a minute
-Thread.sleep(60*1000);
+// let's wait a moment
+Thread.sleep(1000);
 
 // And another round
 readerManager.continueReading();
 
-new Thread(() -> {
-nakadiClient.stream(subscription)
-        .withReaderManager(readerManager)
-        .listen(SalesOrderPlaced.class, listener);  
-}).run();
+// let it run for 1 minute
+Thread.sleep(60*1000);
+
+//And termination
+readerManager.terminateReader();
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -157,6 +157,80 @@ This library is currently tested and used in production with `SimpleClientHttpRe
 
 Please note that `HttpComponentsClientHttpRequestFactory` and also `SimpleClientHttpRequestFactory` since spring 4.3.x try to consume the remaining stream on closing and so might block during reconnection.
 
+## Stopping listening on subscriptions
+
+In order to be able to stop listening on a subscription you need to implement the `ReaderManager` interface.
+
+The interface provides one method `boolean continueReading(Set<String> eventNames, Optional<Subscription> subscription)` which is used by Fahrschein to check if the listening should be stopped.
+
+**Please note** : Resetting the `ReaderManager` is responsibility of the application.
+ 
+```java
+
+public class DemoReaderManager implements ReaderManager {
+  
+  private boolean discontinueReading = false;
+  
+  boolean continueReading(Set<String> eventNames, Optional<Subscription> subscription) {
+    return discontinueReading;
+  }
+  
+  void discontinueReading() {
+    discontinueReading = true;
+  }
+  
+  void continueReading() {
+    discontinueReading = false;
+  }
+}
+
+final String eventName = "sales-order-placed";
+
+// Create a Listener for our event
+final Listener<SalesOrderPlaced> listener = events -> {
+    for (SalesOrderPlaced salesOrderPlaced : events) {
+        LOG.info("Received sales order [{}]", salesOrderPlaced.getSalesOrder().getOrderNumber());
+    }
+};
+
+// Create ReaderManager
+DemoReaderManager readerManager = new DemoReaderManager();
+
+// Configure client, defaults to using the high level api with ManagedCursorManger and SimpleClientHttpRequestFactory
+final NakadiClient nakadiClient = NakadiClient.builder(NAKADI_URI)
+        .withAccessTokenProvider(new ZignAccessTokenProvider())
+        .build();
+
+// Create subscription using the high level api
+Subscription subscriptions = nakadiClient.subscription(applicationName, eventName).subscribe();
+
+// Start streaming, the listen call will block and automatically reconnect on IOException
+new Thread(() -> {
+nakadiClient.stream(subscription)
+        .withReaderManager(readerManager)
+        .listen(SalesOrderPlaced.class, listener);  
+}).run();
+
+// let it run for 1 minute
+Thread.sleep(60*1000);
+
+// and stop the reading
+readerManager.discontinueReading();
+
+// let's wait a minute
+Thread.sleep(60*1000);
+
+// And another round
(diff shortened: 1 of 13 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/StreamBuilder.java
```diff
@@ -47,6 +47,8 @@
 
     StreamBuilder withBackoffStrategy(BackoffStrategy backoffStrategy);
 
+    StreamBuilder withStopSomething(ReaderManager readerManager);
```
tmuehl: `stopSomething`? Maybe, we can find a better name?

fahrschein/src/main/java/org/zalando/fahrschein/ReaderManager.java
```diff
@@ -0,0 +1,10 @@
+package org.zalando.fahrschein;
+
+import org.zalando.fahrschein.domain.Subscription;
+
+import java.util.Optional;
+
+@FunctionalInterface
+public interface ReaderManager {
```
tmuehl: In your first approach the `Listener` interface exposed a method `continueReading`? Why did you separate it now from the `Listener`?

fahrschein/src/main/java/org/zalando/fahrschein/StreamBuilder.java
coders-kitchen: Yup! missed that on a refactoring. is now withReaderManager

coders-kitchen:
After the initial implementation I was thinking if this would not lead to a brittle usage as the user has to remember to reset the listener (somehow). But I think it is the better approach than using the manager I added, as this one introduces more complexity. 

What do you think?

tmuehl:
@coders-kitchen It is indeed a valid consideration that one might easily forget to reset the `Listener`. Depending on the implementation this might also be the case for the `ReaderManager`.

I liked the initial version, but I have no preference to be honest.

@jhorstmann @lukasniemeier-zalando Do you have a preference?

fahrschein/src/main/java/org/zalando/fahrschein/ReaderManager.java
jhorstmann: I'd prefer the separate interface. The current signature only works for the subscription api, to also support the lowlevel api the signature would need to be

    boolean continueReading(Set<String> eventNames);

Or

    boolean continueReading(Set<String> eventNames, Optional<Subscription> subscription);

jhorstmann: Could you also add a default implementation of this interface, similar to the `DefaultErrorHandler` that is used instead of the anonymous functions in `NakadiClientBuilder` and the tests?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -341,6 +342,7 @@ void runInternal() throws IOException, BackoffException {
                 throw e;
             }
         }
+        jsonInput.close();
```
jhorstmann: The closing logic is getting more and more complicated. Normally I would not recommend using exceptions for flow control, but maybe it would simpler to throw a custom exception, similar to how the interrupted flag is handled.

fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderTest.java
```diff
@@ -606,4 +608,29 @@ public void shouldCloseOnRuntimeException() throws IOException {
...
+        final ClientHttpRequest request = mock(ClientHttpRequest.class);
+        when(request.execute()).thenReturn(response);
+        final HttpHeaders headers = new HttpHeaders();
+        when(request.getHeaders()).thenReturn(headers);
+
+        when(clientHttpRequestFactory.createRequest(uri, HttpMethod.GET)).thenReturn(request);
+
+        final NoBackoffStrategy backoffStrategy = new NoBackoffStrategy();
+
+        ReaderManager readerManager = (Optional<Subscription> subscription) -> false;
```
jhorstmann: Should the connection even be opened in this case?

Another testcase would be 

    ReaderManager readerManager = mock(ReaderManager.class);
    when(readerManager.continueReading()).thenReturn(true).thenReturn(false);

And then verify that the listener was called once.

fahrschein/src/main/java/org/zalando/fahrschein/ReaderManager.java
coders-kitchen: Thanks for the comment @jhorstmann.
I'll follow your advice and make the changes you mentioned.

fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderTest.java
coders-kitchen: That's true.

I provide now two test case. one to verify that the reading discontinues when the readermanager flags not to continue reading. the other one, that checks if that the reading is not started at all, if the reader disallows it

fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
coders-kitchen: I extended the README and added an example implementation to the example module

coders-kitchen: I see your point and agree with you, exceptions should not be used for flow control.

But I see the point that the logic gets more and more complicated. I will introduce a custom exception for now, maybe in the future a better way of dealing with it is discovered.

README.md
```diff
@@ -157,6 +157,80 @@ This library is currently tested and used in production with `SimpleClientHttpRe
...
+Thread.sleep(60*1000);
+
+// and stop the reading
+readerManager.discontinueReading();
+
+// let's wait a minute
+Thread.sleep(60*1000);
+
+// And another round
+readerManager.continueReading();
```
jhorstmann: The usecase is much clearer now to me, but I don't see how the current implementation allows resumption of streaming events. After the flag is set to `false` the first time the thread stops, the only way to resume would be to reschedule the `listen` method.

We have an internal library to manage spring thread pools and `@Scheduled` tasks (named `bobcat` in github enterprise), which allows to suspend and resume scheduled tasks using a rest api. I can send you the link or code example on monday. I'm not sure if it tries to `interrupt` currently running tasks, but it should be possible to adapt that library to this usecase.

coders-kitchen: Yes you're right, I also got across this problem.

My idea is to change the ReaderManager in that way that it has two methods.
One is

    boolean discontinueReading(Set<String> eventNames, Optional<Subscription> subscription)

which will be a revertible action. The reader would wait in until it either is terminated or the method return `false`. The wait could be done by a `Thread.sleep(100)`.

The other method would be

    boolean terminateReading(Set<String> eventNames, Optional<Subscription> subscription)

which will not be revertible after it returned `true`. It would be similar to the behaviour of `Thread.currentThread().interrupt()`.

The issue I see at the moment is how to deal with the situation if Nakadi has no free subscription slots left.

I'll push a change to the README to make it a little bit clearer how the idea would look like from a users perspective.

jhorstmann:
I'm still not convinced by the changes and the example. Consider what is already possible without code changes: https://github.com/zalando-incubator/fahrschein/commit/1d1742465a0d9c0a7fd0543fc83cd2b2b156be37#diff-0e7fa37ccfa27ca6f93c8df7bb73ecc4R657

If I'm not misunderstanding your example, this should give the same behaviour.

coders-kitchen:
Sad to hear, but thanks for the input. That's an option I didn't came across yet.

I rethought the whole approach taking your example into consideration and came to the idea, that the NakadiReader shouldn't be aware of the manager at all and the manager is than responsible for accepting listener registrations as well as starting and stopping registered listeners, it could look like this :

    public interface ReaderManager {
      void stopListener(Set<String> eventNames, Optional<Subscription> subscription);

      <T> void registerListener(IORunnable nakadiReader, Set<String> eventNames, Optional<Subscription> subscription);

      void startListener(Set<String> eventNames, Optional<Subscription> subscription) throws IOException;
    }

coders-kitchen:
The last commit was to big, I pushed work in progress code as well as the changed readme.

coders-kitchen:
as discussed closing this PR and create a new one with an example Manager

PR #163, opened 2017-04-13
Stopping and resuming reader with executor

jhorstmann:
Example how #161 can already be achieved with the current code. Introduces new `runnable` method on `StreamBuilder` to simplify using it in a thread pool.


coders-kitchen:
👍

jhorstmann:
:+1:

Issue #164, opened 2017-05-03
Add "unsubscribe" functionality to remove subscriptions from Nakadi


dmitrykrivaltsevich:
With the current version of "fahrschein" library it is possibly to create a subscription. Would be nice to have an "unsubscribe" method to programmatically manage subscriptions. It also enhances a bit a symmetry of a programming interface of this library.

Example use case: having this feature one can remove active subscriptions on shutdown of an application.

jhorstmann:
Fixed by #165

PR #165, opened 2017-05-11
remove subscriptions from Nakadi

AdnanKukuljac:
#164


fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -109,6 +109,26 @@ public SubscriptionBuilder subscription(String applicationName, Set<String> even
...
     }
 
+    /**
+     * Delete subscription based on subscription ID.
+     *
+     * @param subscriptionId
+     * @return
+     * @throws IOException
+     */
+    public boolean deleteSubscription(String subscriptionId) throws IOException {
```
jhorstmann: In all failure cases I would expect an exception, which means this can be a `void` method

jhorstmann:
:+1:

Issue #68, opened 2016-08-15
Document and validate stream parameters


jhorstmann:
- Copy description from nakadi api docs as javadoc
- Validate relation between different timeout parameters

AdnanKukuljac:
@jhorstmann close?

jhorstmann:
Fixed by #170

PR #170, opened 2017-05-18
Stream parameters

AdnanKukuljac:
#68


fahrschein/src/main/java/org/zalando/fahrschein/StreamParameters.java
```diff
@@ -59,22 +60,83 @@ String toQueryString() {
...
         return new StreamParameters(batchLimit, streamLimit, batchFlushTimeout, streamTimeout, streamKeepAliveLimit, maxUncommittedEvents);
     }
 
-    public StreamParameters withStreamLimit(int streamLimit) {
+    /**
+     *
+     * Maximum number of Events in this stream (over all partitions being streamed in this connection).
+     *
+     * @param streamLimit
+     *          If 0 or undefined, will stream batches indefinitely.
```
jhorstmann: This is the issue I remembered, it seems this is not actually true, see https://github.com/zalando-incubator/nakadi-java/issues/125

fahrschein/src/main/java/org/zalando/fahrschein/StreamParameters.java
```diff
@@ -59,22 +62,95 @@ String toQueryString() {
...
+     *          batch_limit must be lower or equal to stream_limit
+     * @return
+     * @throws StreamParametersException
+     */
+    public StreamParameters withBatchLimit(int batchLimit) throws StreamParametersException {
+        if(streamLimit != null && streamLimit < batchLimit){
+            throw new StreamParametersException("streamLimit is lower than batch_limit.");
+        }
+        if(batchLimit < DEFAULT_BATCH_LIMIT){
+            throw new StreamParametersException("batch_limit can't be lower than 1.");
```
jhorstmann: Since these would be programmer errors, I think it should throw a `RuntimeException`, maybe just an `IllegalArgumentException`

jhorstmann:
:+1:

PR #171, opened 2017-05-18
Make ManagedCursorManager extensible

### LINKED ISSUE #169: Support wrapping on ManagedCursorManager
Currently the `NakadiClientBuilder` automagically wraps the `RequestFactory`, providing support for `problem` and adding authorization as defined by the `AccessTokenProvider`.

In case client wants to provide her own `CursorManager`, she has to implement this wrapping on their own as it seems, as the classes are not public.

jhorstmann:
By wrapping the ClientHttpRequestFactory with problem handling and authorization if needed

Fixes #169

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static java.util.Collections.singletonList;
+import static org.zalando.fahrschein.NakadiClientBuilder.wrapClientHttpRequestFactory;
 
 public class ManagedCursorManager implements CursorManager {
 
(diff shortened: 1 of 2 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -73,7 +74,15 @@ void setStreamId(String streamId) {
...
 
+    public ManagedCursorManager(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, AccessTokenProvider accessTokenProvider) {
+        this(baseUri, wrapClientHttpRequestFactory(clientHttpRequestFactory, accessTokenProvider), true);
+    }
+
     public ManagedCursorManager(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory) {
+        this(baseUri, wrapClientHttpRequestFactory(clientHttpRequestFactory, null), true);
+    }
+
+    ManagedCursorManager(URI baseUri, ClientHttpRequestFactory clientHttpRequestFactory, boolean clientHttpRequestFactoryIsAlreadyWrapped) {
```
jhorstmann: Named constructors would be nice :)

lukasniemeier-zalando: Alternatively you could always wrap and change the wrapping code to something like
```
ClientHttpRequestFactory wrapIfNeeded(ClientHttpRequestFactory factory) {
  if (factory instanceof AuthorizedClientHttpRequestFactory) return factory;
  return wrapped(factory);
}
```

This way the wrapping decision would be at one place.

jhorstmann: I would have to check for both `instanceof` `AuthorizedClientHttpRequestFactory` and `ProblemHandlingClientHttpRequestFactory`, the authorization wrapping is only done if there is a ´AccessTokenProvider`.

jhorstmann:
:+1:

PR #175, opened 2017-05-29
Use double-dispatch in multi-event example

jhorstmann:
This moves the processing locic to a central place and allows the usage
of dependencies for handling events.


jhorstmann:
:+1:

PR #177, opened 2017-05-31
Make ProblemHandlingClientHttpRequestFactory and AuthorizedClientHttpRequestFactory public

ivanyu:
`ProblemHandlingClientHttpRequestFactory` and `AuthorizedClientHttpRequestFactory` were made public to allow their usage outside the library. For example, this might be necessary for implementing a custom `CursorManager`, derived from `ManagedCursorManager` (to pass to the constructor).


jhorstmann:
With #171 wrapping should be done automatically when extending `ManagedCursorManager`, does that solve the issue?

ivanyu:
@jhorstmann seems like it, thank you.

PR #179, opened 2017-06-08
Allow wrapping event processing and commit in a transaction

### LINKED ISSUE #178: Transactions around processing of a batch
**As** a client
**I want** to be able to decorate the `void processBatch(Batch<T> batch)` method
**So that** I can have a transaction around it.

See [this method](https://github.com/zalando-incubator/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java#L195). Currently if `cursorManager.onSucess` fails, a client has no chance to roll back whatever she did in the `listener`.

jhorstmann:
Fixes #178

### DIFF
--- a/fahrschein-jdbc/src/main/java/org/zalando/fahrschein/jdbc/TransactionalBatchHandler.java
+++ b/fahrschein-jdbc/src/main/java/org/zalando/fahrschein/jdbc/TransactionalBatchHandler.java
@@ -1,0 +1,16 @@
+package org.zalando.fahrschein.jdbc;
+
+import org.springframework.transaction.annotation.Transactional;
+import org.zalando.fahrschein.IORunnable;
+import org.zalando.fahrschein.BatchHandler;
+
+import java.io.IOException;
+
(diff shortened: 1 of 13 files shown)


fahrschein-jdbc/src/main/java/org/zalando/fahrschein/jdbc/TransactionalBatchHandler.java
```diff
@@ -0,0 +1,16 @@
+package org.zalando.fahrschein.jdbc;
```
lukasniemeier-zalando: Or move to core fahrschein and have `spring-tx` as `provided` dependency?

jhorstmann: I would consider this class more as an example, you might want to customize the rollback behaviour.

jhorstmann:
:+1:

lukasniemeier-zalando:
:+1:

PR #186, opened 2017-07-20
Refactor http api and drop spring dependency

jhorstmann:


fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/HeadersImpl.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+        } else {
+            return readOnly ? Collections.unmodifiableList(list) : list;
+
+        }
+    }
+
+    @Override
+    public void add(String headerName, String value) {
+        get(headerName).add(value);
```
lukasniemeier-zalando: Why can I add stuff if it is `readOnly`

lukasniemeier-zalando:
:+1:

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/HeadersImpl.java
jhorstmann: I'll add some more testcases, but the returned list should be unmodifiable.

jhorstmann: And of course there was a bug in exactly this logic

lukasniemeier-zalando:
:+1:

jhorstmann:
:+1:

PR #195, opened 2017-08-18
Do not ignore status 422 when creating subscription

### LINKED ISSUE #194: Status 422 when trying to subscribe to non-existing event is ignored

jhorstmann:
Fixes #194


jhorstmann:
:+1:

PR #202, opened 2017-09-10
#199 change error handling description in readme

### LINKED ISSUE #199: Documentation is unclear on error handling
README states:

> RuntimeException aborts streaming of events, the user is responsible for handling these

Ok, apparently if my listener throw RuntimeException then stream processing will be aborted.
But what about library code: can it throw RuntimeException as well? 
Should I wrap calls to `subscribe()` and `listen()` into try-catch? This moment is unclear.

I think we can improve docs here.

mkulak:
Closes #199 

If somebody knows how to formulate this better - I'm open for suggestions

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,6 @@
 ## Features
 
  - Consistent error handling
-    - `IOException` handled as retryable
-    - `EventAlreadyProcessedException` is ignored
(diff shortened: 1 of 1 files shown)


jhorstmann:
:+1:

jhorstmann:
Thanks for the contribution. The part about initial IOException is also a bit clearer now.

PR #203, opened 2017-09-26
add dedicated catch clause for InterruptedIOException to prevent race…

coders-kitchen:
… conditions.

As a live situation revealed a race condition between both ``Thread.currentThread().isInterrupted()`` this PR let the main loop explicitly stop when in the try part the thread was interrupted

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
@@ -35,7 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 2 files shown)


jhorstmann:
@coders-kitchen One big problem is that a `SocketTimeoutException` is also an `InterruptedIOException` and we want to retry in that case. Without more information where the race condition is exactly or a testcase I would not want to merge this.

coders-kitchen:
Thanks for the hint.
I replaced the original exception by a dedicated one for this case.

We can observe the race condition very frequently, when we try to shutdown the listener in the way described.

At the moment we work around this by checking in the listener, if it is allowed to be execute. In case not throwing an exception that breaks the execution.

jhorstmann:
But where is the race condition exactly? If this fixes it then something must be resetting the `interrupted` flag between catching the `IOException` and `break`ing from the loop, do you know what this might be?

jhorstmann:
@coders-kitchen can you have a look at #204 and check that it should have the same effect as this PR? Still wondering what might clear the interrupt flag though.

jhorstmann:
@coders-kitchen please reopen if you think [#204](https://github.com/zalando-nakadi/fahrschein/pull/204/files) does not fix he issue

PR #205, opened 2017-10-09
Extract mapping logic into its own interface

jhorstmann:
This allows optimized implementation that only extract a single property
from an event, or for example to map directly to clojure datastructures.


fahrschein/src/main/java/org/zalando/fahrschein/MappingEventReader.java
```diff
@@ -0,0 +1,60 @@
...
+                    events.add(eventClass.cast(eventIterator.next()));
+                } else {
+                    break;
+                }
+            } catch (RuntimeException e) {
+                final Throwable cause = e.getCause();
+                if (cause instanceof JsonMappingException) {
+                    onMappingException((JsonMappingException) cause);
+                } else if (cause instanceof IOException) {
+                    throw (IOException)cause;
```
lukasniemeier-zalando: whitespace :wink:

fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -4,39 +4,31 @@
...
-import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.zalando.fahrschein.domain.Batch;
-import org.zalando.fahrschein.domain.Cursor;
-import org.zalando.fahrschein.domain.Lock;
-import org.zalando.fahrschein.domain.Partition;
-import org.zalando.fahrschein.domain.Subscription;
+import org.zalando.fahrschein.domain.*;
```
lukasniemeier-zalando: on purpose?

fahrschein/src/main/java/org/zalando/fahrschein/JsonParserHelper.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.annotation.Nullable;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InterruptedIOException;
+
+class JsonParserHelper {
+    private JsonParserHelper() {
+    }
+
+    static void expectToken(JsonParser jsonParser, JsonToken expectedToken) throws IOException {
```
lukasniemeier-zalando: does it make sense to mark this as `public`?

jhorstmann: Should be used only internally. Also could not think of a better name than `JsonParserHelper` but for a non-public class this is probably not that important.

jhorstmann:
:+1:

PR #210, opened 2017-11-02
Fixed ZignAccessTokenProvider

### LINKED ISSUE #208: ZignAccessTokenProvider succeeds with empty string on zign command failure
`ZignAccessTokenProvider` does not check the exit code of the `zign` command. Instead returns the empty output of the process as a succcess.

### LINKED ISSUE #209: ZignAccessTokenProvider does not work
I think the `zign` command no longer accepts `uid` as an argument to the `token` subcommand. It should just be `zign token` now.

amrhassan:
There were no existing tests for this part of the code. Do I need to still provide one?

Closes #208, Closes #209

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
@@ -10,6 +10,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 public class ZignAccessTokenProvider implements AccessTokenProvider {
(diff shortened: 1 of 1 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
```diff
@@ -42,18 +43,25 @@ private static String readAll(InputStream inputStream) throws IOException {
...
+            throw new IOException("zign process took longer than 5 seconds to exit");
         }
     }
 
     private static Entry update(@Nullable Entry entry)  {
         final long now = System.currentTimeMillis();
         try {
-            return entry == null || entry.timestamp < now - CACHE_DURATION ? new Entry(now, zign()) : entry;
+            final Entry newEntry = entry == null || entry.timestamp < now - CACHE_DURATION ? new Entry(now, zign()) : entry;
+            LOG.debug("Refreshed token from zign");
```
jhorstmann: This would now log every time, independent of whether the `zign` binary was called.

fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
```diff
@@ -42,18 +43,25 @@ private static String readAll(InputStream inputStream) throws IOException {
...
-        final Process zign = new ProcessBuilder("zign", "token", "uid").start();
+        final Process zign = new ProcessBuilder("zign", "token").start();
         try (final InputStream inputStream = zign.getInputStream()) {
-            return readAll(inputStream).trim();
-        } finally {
-            LOG.debug("Refreshed token from zign");
+            final String output = readAll(inputStream).trim();
+            zign.waitFor(5, TimeUnit.SECONDS);
+            if (zign.exitValue() != 0) {
+                throw new IOException("zign exit code was non-zero");
```
jhorstmann: Could include the exit code in the exception message

jhorstmann:
@amrhassan Thanks for the contribution, I added two small comments, otherwise this looks good.

fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
amrhassan: Sure

fahrschein/src/main/java/org/zalando/fahrschein/ZignAccessTokenProvider.java
amrhassan: Right. Will change.

jhorstmann:
:+1:

PR #212, opened 2017-11-28
Log unexpected throwable before returning from NakadiReader

### LINKED ISSUE #211: Fahrschein thread stopped listening after retrying 12 times
happened in Zalando network yesterday around midnight:

(first retrying 11 times) main cause:
java.net.SocketException: Unexpected end of file from server
(last log from fahrschein library)
o.z.f.ExponentialBackoffStrategy         : Retry [12], sleeping for [64873] milliseconds

after that no more logs from fahrschein. Also thread that was listening was monitored and Future.isDone() was true. How is it possible that thread just stops like this silently instead of retrying further as it said it would?

jhorstmann:
Fixes #211


jhorstmann:
:+1:

PR #220, opened 2018-08-07
Equal & FullJitterBackoffStrategy

ancelotpinto:
Implementation of:
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

- [ ] Add test cases

### DIFF
--- a/fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
+++ b/fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
@@ -21,14 +21,7 @@ import org.apache.http.impl.client.HttpClients;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.client.OkHttp3ClientHttpRequestFactory;
-import org.zalando.fahrschein.EventProcessingException;
-import org.zalando.fahrschein.ExponentialBackoffStrategy;
-import org.zalando.fahrschein.IORunnable;
-import org.zalando.fahrschein.Listener;
-import org.zalando.fahrschein.NakadiClient;
-import org.zalando.fahrschein.NoBackoffStrategy;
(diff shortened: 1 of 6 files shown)


jhorstmann:
Thanks for the PR, this looks good. I currently don't have access to a nakadi instance to test this though, did you already try it against a real instance?

Do you think it makes sense to change the default to `EqualJitterBackoffStrategy`?

ancelotpinto:
Please find the testing details below:
https://gist.github.com/ancelotpinto/15e8859ba198431c40f52e8155b1f036

I believe EqualJitterBackoffStrategy would make for a sane default as is implemented (albeit with some improvements) [here](https://github.com/aws/aws-sdk-java/blob/756d7ac45f053195393170588dc0c8d3ddc9e19f/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java#L143-L156)

I would be happy to make the changes, if we can consider the same for 0.16 release (as this would be a breaking change for clients).

ancelotpinto:
Also to note, the backoff strategy was configured as follows:
new EqualJitterBackoffStrategy(
                                    ExponentialBackoffStrategy.DEFAULT_INITIAL_DELAY,
                                    ExponentialBackoffStrategy.DEFAULT_BACKOFF_FACTOR,
                                    10 * 1000L, -1) and we had:
sleeping for [589]
sleeping for [769]
sleeping for [928]
sleeping for [2166]
sleeping for [2622]
sleeping for [4127]
sleeping for [7134]
sleeping for [5240]
sleeping for [7278]
sleeping for [8147]
sleeping for [9395]
sleeping for [7254]
sleeping for [9947]
sleeping for [6936]
Retry [15], sleeping for [8368] milliseconds

ancelotpinto:
EqualJitter is now the default.
Updated test cases. Please review.

jhorstmann:
:+1:

PR #224, opened 2018-10-02
Add support for Open Tracing's span context.

zekefast:
# Overview
After addition of propagation support of Open Tracing's span context to Nakadi through `span_ctx` attribute in `metadata` we need to be able to reach span context in received events by using client libraries.

Changes in this pull request give ability to retrieve span context data as `Map<String, String>` through event's `getMetadata().getSpanCtx()`.

# Changes
- Introduce new `Metadata` attribute `spanCtx`
- Modify `Metadata` constructors to initialize the attribute
- Add getter for the attribute
- Modify test `NakadiReaderDeserializationTest#shouldDeserializeBusinessEventMetadata` to check for empty span context (`null`) when it absent in JSON payload.
- Add test `NakadiReaderDeserializationTest#shouldDeserializeSpanCtxInBusinessEventMetadata` to check proper deserialization of JSON's span context attributes into `spanCtx` map.

# Related Pull Requests
- https://github.com/zalando/nakadi/pull/931

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.annotation.concurrent.Immutable;
 import java.time.OffsetDateTime;
+import java.util.HashMap;
+import java.util.Map;
 
 @Immutable
 public final class Metadata {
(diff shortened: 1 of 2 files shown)


zekefast:
Do you think it is fine or you would like to see version with [`@JsonAnySetter`](http://www.cowtowncoder.com/blog/archives/2011/07/entry_458.html)?

whiskeysierra:
> Do you think this is fine or would you like to see version with [`@JsonAnySetter`](http://www.cowtowncoder.com/blog/archives/2011/07/entry_458.html)?

I'd prefer a generic extension to the core instead. In addition to a new module that automatically creates a new span around each batch (?) or stream (?).

fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
```diff
@@ -13,23 +15,25 @@
...
     }
 
-    public Metadata(String eventType, String eid, OffsetDateTime occurredAt, OffsetDateTime receivedAt, String flowId) {
+    public Metadata(String eventType, String eid, OffsetDateTime occurredAt, OffsetDateTime receivedAt, String flowId, Map<String, String> spanCtx) {
         this.eventType = eventType;
         this.eid = eid;
         this.occurredAt = occurredAt;
         this.receivedAt = receivedAt;
         this.flowId = flowId;
+        this.spanCtx = spanCtx;
```
jhorstmann: Should we create an immutable copy here? All other properties are immutable.

I'd also suggest adding another constructor without the span, to stay compatible with existing code.

jhorstmann:
@zekefast @whiskeysierra What would be the benefit of `JsonAnySetter` here?

I'm not sure whether the decision to create a span per event or per batch should be done by the library, this probably depends on what the application is doing with the events it receives.

zekefast:
@jhorstmann,

> @zekefast @whiskeysierra What would be the benefit of JsonAnySetter here?
I think @whiskeysierra by using `@JsonAnySetter` would like to avoid interfaces changes for `Metadata` in case of further necessity to extend it with additional attributes, but correct me if I'm wrong.

> I'm not sure whether the decision to create a span per event or per batch should be done by the library, this probably depends on what the application is doing with the events it receives.

I think you are right. Decision to create spans should be very specific to application and domain logic as well as span naming, otherwise there will be almost useless pollution with lots of automatic spans which is make it hard to get you head around once you troubleshooting issues.

@whiskeysierra What would like to see in that additional module (`fahrschein-opentracing`?)? Do we really need automatic span/traces creation? Wouldn't be enough to provide typed interfaces for common fields like traceId and spanId? Do you think about new `interface` like `TracingEvent` for those fields or you would prefer to remove `final` from `Metadata` and extend it to `TracingMetadata`?

whiskeysierra:
> would like to avoid interfaces changes for `Metadata` in case of further necessity to extend it with additional attributes, but correct me if I'm wrong.

Correct.

> What would like to see in that additional module (`fahrschein-opentracing`?)? Do we really need automatic span/traces creation?

Maybe not automatic, but some infrastructure to easily manage spans based on batches/events, e.g. tagging the span with the event type or whatever is useful.

fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
zekefast: @jhorstmann I rolled back changes to constructors and use `JsonAnySetter` to catch span context.

zekefast:
I'm kinda lost :) Could you suggest which way I should go? What's is more preferred immutable embedded in `span_ctx` current `Metadata` class or implementation based on `@JsonAnySetter/@JsonAnyGetter`?

pc-alves:
Adding my 2cent to the matter (and trying to revive this PR =) ).

I feel that the use of `@JsonAnyGetter/@JsonAnySetter` is a bit contradictory looking at the rest of the `Metadata` class. Metadata is non-extensible part of Nakadi events (if I'm not mistaken this is the first change to it). Having said that, a solution that defines some properties of Metadata, but leaves others up to a Map results in an inconsistent solution.
I agree with @jhorstmann that an immutable Map would be the way to go. Consistent with the current implementation, and with the Nakadi specification.

@whiskeysierra On the topic of a module that creates spans per batch or stream that really should be left up to the application owner. A batch can carry events with different span contexts and the decision of which, if any, of those contexts to continue should be on the application owner.
My proposal would be to change the implementation to use an immutable map and leave it at that. As the use of Fahrschein with Distributed Tracing grows we will learn more about how to use it and then we can materialize those learnings into a new module.

jhorstmann:
I'm going to merge this and then refactor to use an immutable map as suggested

jhorstmann:
:+1:

Issue #225, opened 2018-10-23
commit_timeout support for streams


fbrns:
Nakadi will support custom commit_timeouts soon (https://github.com/zalando/nakadi/pull/969). Include commit_timeout as a stream parameter.

jhorstmann:
Fixed by #226

PR #226, opened 2018-10-23
#225 commit_timeout support for streams

fbrns:


unmerge:
:+1:

fahrschein/src/main/java/org/zalando/fahrschein/StreamParameters.java
```diff
@@ -136,7 +143,27 @@ public StreamParameters withStreamTimeout(int streamTimeout) throws IllegalArgum
...
+     *
+     * @param commitTimeout
+     *          Stream initialization will fail if commit_timeout is lower than 0 or higher than 60.
+     * @throws IllegalArgumentException
+     */
+    public StreamParameters withCommitTimeout(int commitTimeout) throws IllegalArgumentException {
+        if(commitTimeout < 0){
+            throw new IllegalArgumentException("commit_timeout can't be lower than 0.");
+        }
+        if(commitTimeout > 60){
```
jhorstmann: I don't think we should be validating the maximum here since the setting on nakadi side seems to be configurable

jhorstmann:
:+1:

PR #227, opened 2018-10-29
Make Metadata and span ctx immutable

jhorstmann:
@zekefast @pc-alves @whiskeysierra please have a quick look

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
@@ -1,14 +1,12 @@
 package org.zalando.fahrschein.domain;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
-import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.annotation.concurrent.Immutable;
 import java.time.OffsetDateTime;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 @Immutable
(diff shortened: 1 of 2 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
```diff
@@ -18,24 +16,29 @@
...
     private final OffsetDateTime receivedAt;
     private final String flowId;
-    private final Map<String, Object> other = new HashMap<String, Object>();
+    private final Map<String, String> spanCtx;
 
     @JsonCreator
     @Deprecated
-    private Metadata(@JsonProperty("event_type") String eventType, @JsonProperty("eid") String eid, @JsonProperty("occurred_at") String occurredAt, @JsonProperty("received_at") String receivedAt, @JsonProperty("flow_id") String flowId) {
-        this(eventType, eid, occurredAt == null ? null : OffsetDateTime.parse(occurredAt), receivedAt == null ? null : OffsetDateTime.parse(receivedAt), flowId);
+    private Metadata(@JsonProperty("event_type") String eventType, @JsonProperty("eid") String eid, @JsonProperty("occurred_at") String occurredAt, @JsonProperty("received_at") String receivedAt, @JsonProperty("flow_id") String flowId, @JsonProperty("span_ctx") Map<String, String> spanCtx) {
```
whiskeysierra: Does it need to be "span_ctx"? Why not "span_context"?

jhorstmann: Good point. On the other hand we already have `eid` instead of event id, so everything is directly matching the names from the nakadi api.

whiskeysierra: Is the property name on nakadi side already decided? @pc-alves

pc-alves: Yes, the change was done using `span_ctx` and not likely to be changed any time soon.

fahrschein/src/test/java/org/zalando/fahrschein/NakadiReaderDeserializationTest.java
```diff
@@ -184,11 +184,11 @@ public void shouldDeserializeBusinessEventMetadata() throws IOException {
         assertThat(metadata.getFlowId(), Matchers.equalTo("ABCD"));
         assertThat(metadata.getOccurredAt(), Matchers.equalTo(OffsetDateTime.of(2016, 10, 26, 19, 20, 21, 123_000_000, ZoneOffset.UTC)));
         assertThat(metadata.getReceivedAt(), Matchers.equalTo(OffsetDateTime.of(2016, 10, 26, 20, 21, 22, 0, ZoneOffset.ofHours(1))));
-        assertThat(metadata.any(), Matchers.equalTo(new HashMap<String, Object>()));
+        assertThat(metadata.getSpanCtx(), Matchers.equalTo(Collections.emptyMap()));
```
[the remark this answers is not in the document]
pc-alves: Not really. Although the majority of open tracing libraries does return a null if no span context is present. I'd say that if you document that an empty map is returned in case of no span context in the metadata we can leave it at that.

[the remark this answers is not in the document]
pc-alves: Actually, turns out that libraries are a bit more diverse than I thought. Some throw exception, others return a null. Returning an empty Map seems like a safe enough approach.

pc-alves:
👍

jhorstmann:
:+1:

PR #230, opened 2019-04-21
#229 add details to ioproblem

### LINKED ISSUE #229: Add details to IOProblem when comitting the cursor fails
When there is an issue with committing the cursor, a the response is [parsed](https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L48) but not always shown in the logs.
Maybe the parsed responseBody can be added as [details](https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/IOProblem.java#L28) (when the reponse does not pass the checks before) in [line 57](https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/IOProblem.java#L28).
As a concrete example, we had an operation that took longer than 60 seconds and so we were over the commit time for the stream. The Response from Nakadi was `org.zalando.fahrschein.IOProblem: Problem [about:blank] with status [422]: [Unprocessable Entity] []` and only via Breakpoint we were able to see the errror message from Nakdi.

Gregsen:
The error response for status code `422` omits the `type` field, which in turn prevented the `detail` field to be shown in the logs.
This PR removes the check for the `type` field (when the response is != 207 || >= 400) and the response is an `application/problem+json`

fixes #229


jhorstmann:
:+1:

PR #235, opened 2019-07-16
#234 - support authorization for subscriptions

### LINKED ISSUE #234: Support authorization for subscriptions
A subscription in Nakadi can have an authorization section. However, it seems that Fahrschein does not yet support it.

Documentation for Nakadi subscriptions: https://nakadi.io/manual.html#definition_Subscription

innokenty:
Fixes #234

This is absolutely a must for environments that don't support creating unauthorized subscriptions.

Tested locally with the Zalando internal Nakadi, works good.

### DIFF
--- a/fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
+++ b/fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
@@ -51,6 +51,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.util.Arrays.asList;
+import static org.zalando.fahrschein.AuthorizationBuilder.authorization;
+import static org.zalando.fahrschein.domain.Authorization.AuthorizationAttribute.ANYONE;
 
 public class Main {
 
@@ -104,6 +106,8 @@ public class Main {
         //simpleListen(objectMapper, listener);
 
         //persistentListen(objectMapper, listener);
+
+        //subscriptionCreateWithAuthorization(objectMapper, listener);
 
         //multiInstanceListen(objectMapper, listener);
     }
@@ -281,6 +285,27 @@ public class Main {
 
         nakadiClient.stream(SALES_ORDER_SERVICE_ORDER_PLACED)
                 .readFromBegin(partitions)
+                .withObjectMapper(objectMapper)
+                .listen(SalesOrderPlaced.class, listener);
+    }
+
+    private static void subscriptionCreateWithAuthorization(ObjectMapper objectMapper, Listener<SalesOrderPlaced> listener) throws IOException {
+
+        final NakadiClient nakadiClient = NakadiClient.builder(NAKADI_URI)
+                .withAccessTokenProvider(new ZignAccessTokenProvider())
+                .build();
+
+        final Subscription subscription = nakadiClient.subscription("fahrschein-demo", SALES_ORDER_SERVICE_ORDER_PLACED)
+                .withConsumerGroup("fahrschein-demo")
+                .withAuthorization(authorization()
+                    .addAdmin("user", "you")
+                    .addAdmin("user", "your_friend")
+                    .addAdmin("user", "your_dog")
+                    .withReaders(ANYONE)
+                    .build())
+                .subscribe();
+
+        nakadiClient.stream(subscription)
                 .withObjectMapper(objectMapper)
                 .listen(SalesOrderPlaced.class, listener);
     }
--- a/fahrschein/src/main/java/org/zalando/fahrschein/AuthorizationBuilder.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/AuthorizationBuilder.java
@@ -3,11 +3,12 @@ package org.zalando.fahrschein;
 import org.zalando.fahrschein.domain.Authorization;
 import org.zalando.fahrschein.domain.Authorization.AuthorizationAttribute;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
-import static java.util.stream.Collectors.toList;
 import static org.zalando.fahrschein.domain.Authorization.AuthorizationAttribute.ANYONE;
 
 /**
@@ -29,7 +30,7 @@ public class AuthorizationBuilder {
      * No restrictions by default, use {@link #withAdmins} / {@link #withReaders} methods to add them.
      */
     public static AuthorizationBuilder authorization() {
-        return new AuthorizationBuilder(singletonList(ANYONE), singletonList(ANYONE));
+        return new AuthorizationBuilder(emptyList(), emptyList());
     }
 
     public AuthorizationBuilder withAdmins(List<AuthorizationAttribute> admins) {
@@ -49,18 +50,22 @@ public class AuthorizationBuilder {
     }
 
     public AuthorizationBuilder addAdmin(String dataType, String value) {
-        List<AuthorizationAttribute> newAdmins = admins.stream().filter(aa -> !aa.equals(ANYONE)).collect(toList());
+        List<AuthorizationAttribute> newAdmins = new ArrayList<>(admins.size() + 1);
+        newAdmins.addAll(admins);
         newAdmins.add(new AuthorizationAttribute(dataType, value));
         return withAdmins(newAdmins);
     }
 
     public AuthorizationBuilder addReader(String dataType, String value) {
-        List<AuthorizationAttribute> newReaders = readers.stream().filter(aa -> !aa.equals(ANYONE)).collect(toList());;
+        List<AuthorizationAttribute> newReaders = new ArrayList<>(readers.size() + 1);
+        newReaders.addAll(readers);
         newReaders.add(new AuthorizationAttribute(dataType, value));
         return withReaders(newReaders);
     }
 
     public Authorization build() {
+        List<AuthorizationAttribute> admins = this.admins.isEmpty() ? singletonList(ANYONE) : this.admins;
+        List<AuthorizationAttribute> readers = this.readers.isEmpty() ? singletonList(ANYONE) : this.readers;
         return new Authorization(admins, readers);
     }
 }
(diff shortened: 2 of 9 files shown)


innokenty:
👍

ePaul:
I'm not really an insider of this library, but the changes look reasonable.

fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
```diff
@@ -285,6 +289,27 @@ private static void persistentListen(ObjectMapper objectMapper, Listener<SalesOr
...
+
+        final Subscription subscription = nakadiClient.subscription("fahrschein-demo", SALES_ORDER_SERVICE_ORDER_PLACED)
+                .withConsumerGroup("fahrschein-demo")
+                .withAuthorization(authorization()
+                    .addAdmin("user", "you")
+                    .addAdmin("user", "your_friend")
+                    .addAdmin("user", "your_dog")
+                    .withReaders(ANYONE)
+                    .build())
+                .subscribe();
```
innokenty: Here's a usage example.

fahrschein/src/main/java/org/zalando/fahrschein/AuthorizationBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static java.util.Arrays.asList;
+
+public class AuthorizationBuilder {
+
+    private final List<AuthorizationAttribute> admins;
+    private final List<AuthorizationAttribute> readers;
+
+    private AuthorizationBuilder() {
+        admins = null;
```
whiskeysierra: Shouldn't those be empty lists?

innokenty: I am afraid that with the empty lists you will get "zero admins" instead of "authorization not set". Maybe Nakadi API experts can clarify?

innokenty: Tried it, you're right!
```
[422]: [Unprocessable Entity] [Field "authorization.readers" may not be null
```

Whole `authorization` can be `null` (meaning no restrictions), but **if** it's there — then all the fields have to be set.

Will fix.

innokenty: Hm, but actually.... it can't be an empty array either!
```
[422]: [Unprocessable Entity] [Field "authorization.readers" must contain at least one attribute
```

What do you think: should I just rewrite the builder then, so it's impossible to create an instance no providing all the needed info? Or do we want to leave it like that and just let clients get errors from the API if they use it wrong? I think it's better to rewrite not to allow improper requests at all.

innokenty: Or maybe just set defaults to `ANYONE`? Guys, what do you think? This is a matter of opinions.

innokenty: Implemented the behaviour that it's `*/*` by default and you can only override it with builder methods. Added a test and javadoc.

whiskeysierra: > Implemented the behaviour that it's `*/*` by default and you can only override it with builder methods. Added a test and javadoc.

Sounds good.

fahrschein/src/main/java/org/zalando/fahrschein/domain/Authorization.java
```diff
@@ -2,6 +2,7 @@
 
 import javax.annotation.concurrent.Immutable;
 import java.util.List;
+import java.util.Objects;
 
 @Immutable
 public class Authorization {
```
whiskeysierra: `final`, otherwise it's not immutable

fahrschein/src/main/java/org/zalando/fahrschein/domain/Authorization.java
```diff
@@ -0,0 +1,63 @@
...
+        public int hashCode() {
+            return Objects.hash(dataType, value);
+        }
+    }
+
+    private final List<AuthorizationAttribute> admins;
+    private final List<AuthorizationAttribute> readers;
+
+    public Authorization(List<AuthorizationAttribute> admins, List<AuthorizationAttribute> readers) {
+        this.admins = admins;
```
whiskeysierra: Should make defensive copies if you want to be immutable.

whiskeysierra: Or use persistent, immutable collections https://github.com/GlenKPeterson/Paguro without exposing them.

innokenty: Yeah, I changed to `java.util.Collections.unmodifiableList`, because this is how it's done in other domain objects in this package already.

fahrschein/src/main/java/org/zalando/fahrschein/domain/Authorization.java
```diff
@@ -0,0 +1,65 @@
...
+        public int hashCode() {
+            return Objects.hash(dataType, value);
+        }
+    }
+
+    private final List<AuthorizationAttribute> admins;
+    private final List<AuthorizationAttribute> readers;
+
+    public Authorization(List<AuthorizationAttribute> admins, List<AuthorizationAttribute> readers) {
+        this.admins = unmodifiableList(admins);
```
whiskeysierra: This just ensure that `this.admins` can't be modified later. But the originally passed list `admins` could still change.

fahrschein/src/main/java/org/zalando/fahrschein/AuthorizationBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public AuthorizationBuilder addAdmin(String dataType, String value) {
+        List<AuthorizationAttribute> newAdmins = admins.stream().filter(aa -> !aa.equals(ANYONE)).collect(toList());
+        newAdmins.add(new AuthorizationAttribute(dataType, value));
+        return withAdmins(newAdmins);
+    }
+
+    public AuthorizationBuilder addReader(String dataType, String value) {
+        List<AuthorizationAttribute> newReaders = readers.stream().filter(aa -> !aa.equals(ANYONE)).collect(toList());;
```
whiskeysierra: I'd do this in the getter.

1. Initialize the lists with an empty list
2. Add admin/reader by creating a copy of the current list + the new role
3. In getter
   - if list is empty, return `singletonList(ANYONE)`
   - else, return admins/readers

whiskeysierra: If you don't want to make a full copy every time, I'd suggest to use https://github.com/GlenKPeterson/Paguro.

whiskeysierra: `return withReaders(admins.append(new AuthorizationAttribute(..)));`

fahrschein/src/main/java/org/zalando/fahrschein/domain/Authorization.java
innokenty: True, but I don't see any problem about it. This is how it's done in the rest of the project, I don't really wanna change the approach for one DTO only.

fahrschein/src/main/java/org/zalando/fahrschein/AuthorizationBuilder.java
innokenty: Yeah, sounds like a good idea, let me try. I am not sure it will be more elegant than the current state, but we'll see.

innokenty: Implemented your proposal, see below. It got 4 lines bigger and arguably better. I'm fine with whatever approach, I think the one you suggested is also quite nice.

I don't wanna add any dependencies just for the sake of elementary collections mapping, because this is a library, that's used across many projects and dragging more dependencies to all of them sounds like a bad idea to me.

whiskeysierra:
:+1:

innokenty:
👍

innokenty:
Great, should we merge & release? Also can you?

whiskeysierra:
Just requested access rights internally.

fahrschein/src/main/java/org/zalando/fahrschein/domain/Authorization.java
jhorstmann: Should be fine here since the builder always creates new lists, so it's impossible to modify it from outside the library

jhorstmann:
:+1:

jhorstmann:
@whiskeysierra let me know if you have any questions about the release script

innokenty:
@jhorstmann thanks for review, we both got access rights now internally. I am willing to merge this and release soon. Should I just run `release.sh <new version>`? I am not sure if I have correct rights for maven publishing....

jhorstmann:
The release process still has some manual steps, for example with 0.18.0 as the next released version after this pr is merged:

 - `git checkout -b release/0.18.0`
 - ./release.sh 0.18.0 0.19.0-SNAPSHOT
 - if the above fails because of missing maven permissions you have to reset the branch to master and also delete the already created 0.18.0 tag before retrying again
 - `git push`
 - open a pr from the release branch to master to update the snapshot version in master
 - `git push --tags`

innokenty:
Ok, that's easy, but what maven permissions do I need? At the moment I'm just a random guy on the internet, how is it checking that not everyone is able to deploy a new version? Is maven repo linked to this github repo somehow, so being a collaborator is enough?

jhorstmann:
As far as I remember, you need a sonatype account which can then be linked or invited to the org.zalando group. I don't know who is currently managing that. You also need a gpg key when publishing, but I think this is only used (optionally) when downloading releases, not for verifying your publishing permissions.

See <https://central.sonatype.org/pages/ossrh-guide.html>, the second step there should only be needed when you want to claim your own group id.

innokenty:
Sadly the squash & merge feature is disabled for this project, so I have to just merge with 4 commits. If someone has enough rights to enable it — please consider.

jhorstmann:
I disabled squash on purpose, that ensures that the actual commit that is tagged as released is part of the master branch and we can be sure that all released features are part of the development branch.

innokenty:
Alright, that's a matter of preference.
Release published, by the way.
And side note: gpg key **is** needed.

PR #236, opened 2019-07-18
Allow any type of authorization type

lukasniemeier-zalando:
Right now the `AuthorizedRequestFactory` is hardwired to `Bearer` authorization, by prefixing it to whatever the `AccessTokenProvider` provides. In this change we introduce a new `AuthorizationProvider` interface providing the whole authorization string, allowing users to choose freely.

* add new AuthorizationProvider interface providing full authorization string
* add adapter for AccessTokenProvider
* extend public interfaces with new provider

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
@@ -11,7 +11,6 @@ import org.zalando.fahrschein.http.api.ContentType;
 import org.zalando.fahrschein.http.api.Request;
 import org.zalando.fahrschein.http.api.RequestFactory;
 import org.zalando.fahrschein.http.api.Response;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -21,8 +20,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import static java.util.Collections.singletonList;
 import static org.zalando.fahrschein.NakadiClientBuilder.wrapClientHttpRequestFactory;
+import static java.util.Collections.singletonList;
 
 public class ManagedCursorManager implements CursorManager {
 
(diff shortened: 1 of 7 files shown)


jhorstmann:
How about having the `AccessTokenProvider` extend the new interface using default methods?

```
public interface AuthorizationProvider {
    String getAuthorizationHeader() throws IOException;
}

public interface AccessTokenProvider extends AuthorizationProvider {
    String getAccessToken() throws IOException;

    default String getAuthorizationHeader() throws IOException {
        return "Bearer ".concat(getAccessToken());
    }
}
```

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -73,8 +72,8 @@ void setStreamId(String streamId) {
     private final ObjectMapper objectMapper;
     private final Map<String, SubscriptionStream> streams;
 
-    public ManagedCursorManager(URI baseUri, RequestFactory clientHttpRequestFactory, AccessTokenProvider accessTokenProvider) {
-        this(baseUri, wrapClientHttpRequestFactory(clientHttpRequestFactory, accessTokenProvider), true);
+    public ManagedCursorManager(URI baseUri, RequestFactory clientHttpRequestFactory, AuthorizationProvider authorizationProvider) {
```
jhorstmann: If you want to be paranoid about binary compatibility there should be a separate overload. Could lead to problems when you have a transitive dependency compiled against the older version, however unlikely that is.

lukasniemeier-zalando: I'd take the risk her, fahrschein is on `0.x.x` either way in the end :wink:

lukasniemeier-zalando:
:+1:

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
whiskeysierra: > transitive dependency compiled against the older version

Fahrschein is not really used in other libraries, is it?

jhorstmann: Github is telling me it is used by [nakadi-producer-spring-boot-starter](https://github.com/zalando-nakadi/nakadi-producer-spring-boot-starter) and [fahrschein-spring-boot-starter](https://github.com/jbellmann/fahrschein-spring-boot-starter). There might be others that are non-public. Anyway, it should fail rather loudly during testing if there is an issue.

innokenty: These two 👆 are mostly authored by @ePaul

PR #240, opened 2019-12-19
toString method for MetaData

AlGreed:
* added toString method for MetaData to get a string representation of such objects in logs
* added "dist: trusty" to .travis.yml to be able to use oraclejdk8 further. Otherwise we need to change jdk  to oraclejdk8, for example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 jdk:
   - oraclejdk8
 install: true
+dist: trusty
 env:
   matrix:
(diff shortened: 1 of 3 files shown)


iav0207:
👍

iav0207:
@whiskeysierra @jhorstmann 
Can any of you please review this tiny PR?

whiskeysierra:
:+1:

AlGreed:
@whiskeysierra Could you please merge it? And i will prepare a PR with release version.

jhorstmann:
Normally I would advise against putting larger toString outputs into log messages and instead log only specified fields, but a toString also helps during debugging, so LGTM.

PR #251, opened 2021-02-01
Exception while the cursor is being committed

### LINKED ISSUE #250: Exception while the cursor is being committed
A `NullPointerException` is thrown while committing the cursor after messages have been processed.

```
java.lang.NullPointerException: null
	at org.zalando.fahrschein.ProblemHandlingRequest.mightBeProblematic(ProblemHandlingRequest.java:74)
	at org.zalando.fahrschein.ProblemHandlingRequest.execute(ProblemHandlingRequest.java:44)
	at org.zalando.fahrschein.ManagedCursorManager.onSuccess(ManagedCursorManager.java:124)
	at org.zalando.fahrschein.NakadiReader$2.run(NakadiReader.java:210)
	at org.zalando.fahrschein.DefaultBatchHandler.processBatch(DefaultBatchHandler.java:13)
	at org.zalando.fahrschein.NakadiReader.processBatch(NakadiReader.java:205)
	at org.zalando.fahrschein.NakadiReader.readBatch(NakadiReader.java:400)
	at org.zalando.fahrschein.NakadiReader.runInternal(NakadiReader.java:290)
	at org.zalando.fahrschein.NakadiReader.run(NakadiReader.java:266)
	at o.z.f.StreamBuilders$AbstractStreamBuilder.listen(StreamBuilders.java:57)
	...
```

The stacktrace indicates that the NPE is thrown after trying accessing the `ContentType` of the response received from Nakadi ([here](https://github.com/zalando-nakadi/fahrschein/blob/b2f0ab4d86a2ac286d99933682033bd5c468ce2b/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L74)).

a-liut:
close #250.

### Solution

The solution adopted in this PR is to allow `fahrschein` to parse the content of the response from Nakadi even if the response ContentType is null.

With this fix, the `NullPointerException` is avoided. A different exception might be thrown during the parsing but such exception will be caught by the surrounding try-catch as usual.

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
@@ -41,7 +41,7 @@ class ProblemHandlingRequest implements Request {
                 final Headers headers = response.getHeaders();
                 final ContentType contentType = headers.getContentType();
 
-                if (mightBeProblematic(contentType)) {
+                if (contentType == null || mightBeProblematic(contentType)) {
 
                     final JsonNode json = objectMapper.readTree(response.getBody());
 


dr0pi:
👍

jhorstmann:
Under which circumstances does nakadi actually return a problem response without content type header?

MALPI:
hey @a-liut @dr0pi,

do you have a response to the question raised by @jhorstmann. Is this change still required?

@lmontrieux do you have any opinion on that?

lmontrieux:
@MALPI nakadi shouldn't do that, but I can't guarantee there isn't a bug somewhere. @a-liut where did you encounter this?

a-liut:
hey, sorry for the late reply.

I'm actually not sure about the reason why and when this situation occurs.

I believe the reason is related to how Nakadi is deployed or on the underlying broker Kafka.

> where did you encounter this?

I posted a stack trace in issue #250. It was thrown when after the cursors have been committed while accessing the response received from Nakadi.

otrosien:
I can reproduce it when getting a `401 Unauthorized` - then we also don't get any content-type back.

otrosien:
I would say the code is not worse-off with the PR in, even if Nakadi should not behave like this, but with the nakadi-docker end2end tests I wasn't able to trigger such an error. Any objections to just merge this @MALPI @othon2000 ?

MALPI:
👍

PR #253, opened 2021-08-30
Log 'info.debug' from 'Batch' in NakadiReader instead of skipping it

sameei:
'Nakadi' uses 'info' section in each 'Batch' to give a hint about the state of the stream. 
Current implementation of 'NakadiReader' ignore it completely, while in our experience we found that those hints help to debug consuming problems. This PR includes small changes to log the 'info.debug' value instead of skipping it.


MALPI:
Hey there,

we just kicked off a meeting in the round of new maintainers taking care of the project.

As part of that we decided that we are going to close the pull requests that are open since a long time since we don't know if these features are still required.

In case the feature is still required don't hesitate to re-open and reach out to us.

Thanks a lot for your contribution!

Malte

Issue #254, opened 2021-11-18
Offset commit timeout errors give no hints in the logs


othon2000:
(PR will follow)

According to the current Nakadi API:
```
        Endpoint for committing offsets of the subscription. If there is uncommited data, and no commits happen
        for 60 seconds, then Nakadi will consider the client to be gone, and will close the connection. As long
        as no events are sent, the client does not need to commit.
        If the connection is closed, the client has 60 seconds to commit the events it received, from the moment
        they were sent. After that, the connection will be considered closed, and it will not be possible to do
        commit with that `X-Nakadi-StreamId` anymore.
        When a batch is committed that also automatically commits all previous batches that were
        sent in a stream for this partition.
```
(source) https://github.com/zalando/nakadi/blob/e83f5f045c75a1bb3f37a74ab3a874b1e197529b/docs/_data/nakadi-event-bus-api.yaml#L1318

If we try to commit a cursor offset after 60 seconds, it will fail. In such cases, the following error message is produced by Fahrschein:
```
org.zalando.fahrschein.NakadiReader      : Exception while processing events for [xxx] on partition [3] at offset [yyy]
org.zalando.fahrschein.IOProblem: Problem [about:blank] with status [422]: [Unprocessable Entity] [Session with stream id zzz not found]
	at org.zalando.fahrschein.ProblemHandlingRequest.handleProblem(ProblemHandlingRequest.java:105) ~[fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.ProblemHandlingRequest.execute(ProblemHandlingRequest.java:53) ~[fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.ManagedCursorManager.onSuccess(ManagedCursorManager.java:124) ~[fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.NakadiReader$2.run(NakadiReader.java:210) ~[fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.DefaultBatchHandler.processBatch(DefaultBatchHandler.java:13) [fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.NakadiReader.processBatch(NakadiReader.java:205) [fahrschein-0.18.0.jar!/:na]
	at org.zalando.fahrschein.NakadiReader.readBatch(NakadiReader.java:400) [fahrschein-0.18.0.jar!/:na]…(more)
```

[...]

PR #255, opened 2021-11-19
Issues/254 commit offset error msg

othon2000:
Addressing https://github.com/zalando-nakadi/fahrschein/issues/254

For cursor offset commit timeouts, instead of:
```
org.zalando.fahrschein.IOProblem: Problem [about:blank] with status [422]: [Unprocessable Entity] [Session with stream id zzz not found]
```
we will get
```
org.zalando.fahrschein.CursorOffsetCommitException: Cursor for subscription [sss] to event [eee] in partition [p] with offset [ooo] failed to commit because of error 422 (Unprocessable Entity). This likely means that the processing time of the batch exceeded the timeout of 60 seconds. In such case, you may want to investigate the slowness in the processing, and/or reduce the batch size
org.zalando.fahrschein.IOProblem: Problem [about:blank] with status [422]: [Unprocessable Entity] [Session with stream id zzz not found]
``` 

Some remarks: I decided to not add more logic to ProblemHandlingRequest ([example](https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L82)), because parsing the response body to determine the error type is far from ideal. Instead, we catch and re-throw a specific exception where the error occurs.

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
@@ -31,7 +31,7 @@ public class CursorOffsetCommitException extends IOException {
             case 422:
                 msg = String.format(
                     "Cursor for subscription [%s] to event [%s] in partition [%s] with offset [%s] failed to commit because of error 422 (Unprocessable Entity). " +
-                    "This likely means that the processing time of the batch exceeded the timeout of 60 seconds. " + 
+                    "This likely means that the processing time of the batch exceeded the timeout (defaults to 60 seconds). " + 
                     "In such case, you may want to investigate the slowness in the processing, and/or reduce the batch size.", 
                     nullSafe(subscriptionId), nullSafe(cursor.getEventType()), cursor.getPartition(), cursor.getOffset());
                 break;
--- a/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
@@ -11,6 +11,13 @@ public class CursorOffsetCommitException extends IOException {
     private final Cursor cursor;
     private final String subscriptionId;
 
+    public CursorOffsetCommitException(int statusCode, Cursor cursor, String subscriptionId, IOException cause) {
+        super(formatMessage(statusCode, cursor, subscriptionId), cause);
+        this.statusCode = statusCode;
+        this.cursor = cursor;
+        this.subscriptionId = subscriptionId;
+    }
+
     public CursorOffsetCommitException(int statusCode, Cursor cursor, String subscriptionId) {
         super(formatMessage(statusCode, cursor, subscriptionId));
         this.statusCode = statusCode;
--- a/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
@@ -1,0 +1,51 @@
+package org.zalando.fahrschein;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import org.zalando.fahrschein.domain.Cursor;
+
+@SuppressWarnings("serial")
+public class CursorOffsetCommitException extends IOException {
+    private final int statusCode;
+    private final Cursor cursor;
+    private final String subscriptionId;
+
+    public CursorOffsetCommitException(int statusCode, Cursor cursor, String subscriptionId) {
+        super(formatMessage(statusCode, cursor, subscriptionId));
+        this.statusCode = statusCode;
+        this.cursor = cursor;
+        this.subscriptionId = subscriptionId;
+    }
+
+    private static String formatMessage(int statusCode, Cursor cursor, String subscriptionId) {
+        String msg;
+        switch (statusCode) {
+            case 422:
+                msg = String.format("Cursor for subscription [%s] to event [%s] in partition [%s] with offset [%s] failed to commit because of error 422 (Unprocessable Entity). This likely means that the processing time of the batch exceeded the timeout of 60 seconds. In such case, you may want to investigate the slowness in the processing, and/or reduce the batch size", 
+                                    nullSafe(subscriptionId), nullSafe(cursor.getEventType()), cursor.getPartition(), cursor.getOffset());
+                break;
+            default: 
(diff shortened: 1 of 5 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
```diff
@@ -0,0 +1,58 @@
...
+        this.statusCode = statusCode;
+        this.cursor = cursor;
+        this.subscriptionId = subscriptionId;
+    }
+
+    private static String formatMessage(int statusCode, Cursor cursor, String subscriptionId) {
+        String msg;
+        switch (statusCode) {
+            case 422:
+                msg = String.format("Cursor for subscription [%s] to event [%s] in partition [%s] with offset [%s] failed to commit because of error 422 (Unprocessable Entity). This likely means that the processing time of the batch exceeded the timeout of 60 seconds. In such case, you may want to investigate the slowness in the processing, and/or reduce the batch size",
```
ePaul, line 32: Could you break that line so no horizontal scrolling is needed in Github?

ePaul: (In some cases, the user might have set the commit timeout smaller than 60 seconds, so I'm not sure this needs to be mentioned in the text.)

othon2000: adjusted

ePaul:
:+1:

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -129,8 +129,12 @@ public void onSuccess(String eventName, Cursor cursor) throws IOException {
             } else if (status == 200) {
                 LOG.warn("Cursor for subscription [{}] to event [{}] in partition [{}] with offset [{}] was already committed", subscriptionId, eventName, cursor.getPartition(), cursor.getOffset());
             } else {
-                throw new IOException(String.format("Unexpected status code [%s] for subscription [%s] to event [%s]", status, subscriptionId, eventName));
+                throw new CursorOffsetCommitException(status, cursor, subscriptionId);
             }
+        } catch (IOProblem e) {
+            // in order to not have to refactor the entire ProblemHandlingRequest class,
+            // we are catching the error here and re-throwing a more specific one
+            throw new CursorOffsetCommitException(e.getStatusCode(), cursor, subscriptionId, e);
```
adyach, line 137: worth mentioning exception in the method signature, because at the moment it worths IOException, but I do not see where exactly it is thrown

fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
```diff
@@ -0,0 +1,61 @@
...
+        this.cursor = cursor;
+        this.subscriptionId = subscriptionId;
+    }
+
+    private static String formatMessage(int statusCode, Cursor cursor, String subscriptionId) {
+        String msg;
+        switch (statusCode) {
+            case 422:
+                msg = String.format(
+                    "Cursor for subscription [%s] to event [%s] in partition [%s] with offset [%s] failed to commit because of error 422 (Unprocessable Entity). " +
```
adyach, line 33:
```suggestion
                    "Cursor for subscription [%s] and event type [%s] in partition [%s] with offset [%s] failed to commit because of error 422 (Unprocessable Entity). " +
```

adyach:
👍

fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
```diff
@@ -0,0 +1,61 @@
+package org.zalando.fahrschein;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import org.zalando.fahrschein.domain.Cursor;
+
+@SuppressWarnings("serial")
+public class CursorOffsetCommitException extends IOException {
```
antban, line 9: It makes sense change the name to `CommitException` (or `CursorCommitException`), as right now class name is incorrectly representing what have happened.

fahrschein/src/main/java/org/zalando/fahrschein/CursorOffsetCommitException.java
antban, line 33: error -> status code (or HTTP status code)

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -129,8 +129,12 @@ public void onSuccess(String eventName, Cursor cursor) throws IOException {
             } else if (status == 200) {
                 LOG.warn("Cursor for subscription [{}] to event [{}] in partition [{}] with offset [{}] was already committed", subscriptionId, eventName, cursor.getPartition(), cursor.getOffset());
             } else {
-                throw new IOException(String.format("Unexpected status code [%s] for subscription [%s] to event [%s]", status, subscriptionId, eventName));
+                throw new CursorOffsetCommitException(status, cursor, subscriptionId);
```
antban, line 132: It actually make sense to try and read nakadi response (and expose it to a user), as the information about problem may be written there.

antban (commented):
Nice! Let's improve it a bit and merge

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
antban: Regarding parsing - I meant to try and extract string representation, not to actually do string parsing aferwards

othon2000: Done. Note that when we catch the `IOProblem` we already have the response body, I've just propagated the root cause. When we throw the exception based on status code, I'm getting the body and adding it to the exception, which also logs it.

othon2000:
All comments addressed.

jhorstmann:
As you might have noticed from the other open PRs, fahrschein is not really maintained at the moment, or rather looking for a new maintainer. I can still merge PRs, but have no real way of testing and probably also can't deploy to the org.zalando groupid on maven central anymore. If someone wants to take over as a maintainer that would be great and I'd be happy to help with any issues in the release script.

othon2000:
@jhorstmann please merge the PR. I'm looking for a new maintainer within Z, and may also become one if needed. Will update you soon.

jhorstmann:
:+1:

PR #284, opened 2022-01-25
Release 0.19.0

MALPI:
Set proper release version

### DIFF
--- a/fahrschein-example/pom.xml
+++ b/fahrschein-example/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>org.zalando</groupId>
         <artifactId>fahrschein-parent</artifactId>
-        <version>0.20.0-SNAPSHOT</version>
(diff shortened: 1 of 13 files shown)


MALPI:
:+1:

otrosien:
:+1:

fahrschein-example/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>org.zalando</groupId>
         <artifactId>fahrschein-parent</artifactId>
-        <version>0.20.0-SNAPSHOT</version>
+        <version>0.19.0</version>
```
othon2000, line 10: I think we can remove the `<version>` tag from all submodules, and they will inherit from the parent.

MALPI: Lets consider an extra issue for that.

PR #293, opened 2022-02-07
ISSUE-259 Set the mandatory paramter pointing ot the right url

MALPI:
Re-introduce nexus url parameter again pointing to the right address


otrosien:
👍

PR #294, opened 2022-02-07
Deprecate external Cursor management

otrosien:
Nakadi's low-level API has one use-case: if you always want to read a
topic from the beginning. This is incompatible with the idea of an
external persistant storage for offsets. If offset persistance is
needed, people should use the normal subscription APIs.

This commit marks the JDBC and Redis implementations as deprecated, so
that we can remove them in a future release.

Signed-off-by: Oliver <oliver.trosien@zalando.de>


MALPI:
:+1:

othon2000:
:+1:

PR #311, opened 2022-03-31
Support ZStandard compression

### LINKED ISSUE #310: Support ZStd compression
Zstd compression is supported by Nakadi, and preferred for small payloads, e.g. when publishing events.

otrosien:
Adds ZStandard compression support, based on the same library that also Nakadi is using.

* Added compression support for both GZip and ZStandard
* Parametrized e2e-tests to check for all variations (NakadiClientE2ETest)
* Removed duplications of RequestFactory tests into abstract base class (AbstractRequestFactoryTest)
* Limits compression to `POST` requests, as [Nakadi only supports it for those](https://github.com/zalando/nakadi/blob/master/core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java#L52). Luckily Fahrschein does not send any PATCH or PUT anyway. 
* cleaned up "NOP logger" warnings from slf4j if no logger implementation was given in test scope
* Introduces "fahrschein-core" module to be able to end-to-end test all http client implementations without maven complaining about cyclic dependencies. (https://github.com/zalando-nakadi/fahrschein/issues/318)

Closes #310 

## Tasks

* [x] Code implementation & unit tests
* [x] Cleanup of duplications in test code
* [x] End-to-end test 
* [x] README documentation


otrosien:
moving the subproject dependency problem into https://github.com/zalando-nakadi/fahrschein/issues/318

otrosien:
re-opening after rebasing on #319

PR #315, opened 2022-04-04
Add documentation

### LINKED ISSUE #131: Could we have some Javadoc in the public API, please?
This library needs some Java documentation comments in the public API classes and methods.

Having some examples of the usage is good, but not a replacement for a documentation in the classes and methods which users are supposed to use. Those will also show up in IDEs, greatly helping in using the API.

I'm here thinking of the NakadiClient class, its Builder, the Subscription class, etc. – but I guess you know better which classes I'm supposed to use.

### LINKED ISSUE #241: [documentation] Documentation for NakadiClient.publish is missing
The publish function should be documented, especially as it is not obvious, a camelCase -> snake_case conversion is applied on the events.

MALPI:
Start documenting public classes and methods with Javadoc

Closes #241 , closes #131

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
@@ -6,7 +6,9 @@ import java.io.IOException;
 import java.util.Formatter;
 import java.util.Locale;
 
-@SuppressWarnings("serial")
+/**
+ * Thrown in case the client wasn't able to publish the given Event to Nakadi.
+ */
 public class EventPublishingException extends IOException {
     private final BatchItemResponse[] responses;
 
--- a/fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
@@ -7,7 +7,9 @@ import java.util.Formatter;
 import java.util.Locale;
 
 /**
- * Thrown in case the client wasn't able to publish the given event to Nakadi.
+ * Thrown in case the client wasn't able to publish the given batch of events to Nakadi.
+ *
+ * The response will contain an array of {@code BatchItemResponse}.
  */
 public class EventPublishingException extends IOException {
     private final BatchItemResponse[] responses;
--- a/fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
@@ -81,8 +81,8 @@ public class NakadiClient {
      * @param eventName where the event should be written to
      * @param events that should be written
      * @param <T> Type of the Event
-     * @throws IOException in case we fail reaching Nakadi or we are unable to write the event.
-     * @throws EventPublishingException in case we fail reaching Nakadi or we are unable to write the event.
+     * @throws IOException in case we fail to reach Nakadi
+     * @throws EventPublishingException In case Nakadi returns an Erroneous response
      */
     public <T> void publish(String eventName, List<T> events) throws EventPublishingException, IOException {
         final URI uri = baseUri.resolve(String.format("/event-types/%s/events", eventName));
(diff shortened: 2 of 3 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
```diff
@@ -6,7 +6,9 @@
 import java.util.Formatter;
 import java.util.Locale;
 
-@SuppressWarnings("serial")
+/**
+ * Thrown in case the client wasn't able to publish the given Event to Nakadi.
```
otrosien, line 10:
```suggestion
 * Thrown in case the client wasn't able to publish the given event to Nakadi.
```

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -38,6 +41,14 @@
     private final ObjectMapper objectMapper;
     private final CursorManager cursorManager;
 
+    /**
+     * Returns a new Builder that will make use of the given request Factory and tries to connect to the give
```
otrosien, line 45:
```suggestion
     * Returns a new Builder that will make use of the given RequestFactory and tries to connect to the give
```

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -60,7 +77,14 @@ public static NakadiClientBuilder builder(URI baseUri, RequestFactory clientHttp
         }
     }
 
-    public <T> void publish(String eventName, List<T> events) throws EventPublishingException, IOException {
+    /**
+     * Writes the given Events to the endpoint provided by the eventName.
```
otrosien, line 81:
```suggestion
     * Writes the given events to the endpoint provided by the eventName.
```

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
otrosien: It doesn't straight away try to connect to the given address...

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -38,6 +41,14 @@
...
 
+    /**
+     * Returns a new Builder that will make use of the given request Factory and tries to connect to the give
+     * address.
+     *
+     * @param baseUri that we try to connect to
+     * @param clientHttpRequestFactory that we use for the execution of our HTTP Requests.
+     * @return A builder to initialize the client. Can be further modified later.
+     */
     public static NakadiClientBuilder builder(URI baseUri, RequestFactory clientHttpRequestFactory) {
```
otrosien, line 52: as part of this PR simplify the var naming to `requestFactory`?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -60,7 +77,14 @@ public List<Partition> getPartitions(String eventName) throws IOException {
         }
     }
 
-    public <T> void publish(String eventName, List<T> events) throws EventPublishingException, IOException {
```
otrosien, line 63: does it make sense to keep this EventPublishingException part of the API signature (even though included in IOException) because it may be treated differently?

MALPI: I've re-added it for now. But to be honest I don't see why it should be treated different?

otrosien: The javadoc even suggests that they're identical. Can we differentiate this somehow? So when is EventPublishingException thrown, to my knowledge not when we fail reaching Nakadi but when Nakadi responds with a partial/complete failure. Document that, and then state IOException for everything else (mostly Nakadi not reachable).

And regarding different treatment: An EventPublishingException contains the failed event ids, so potentially you can recover and re-send the failed events only.

fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingException.java
```diff
@@ -6,7 +6,9 @@
 import java.util.Formatter;
 import java.util.Locale;
 
-@SuppressWarnings("serial")
+/**
+ * Thrown in case the client wasn't able to publish the given event to Nakadi.
```
otrosien, line 10: batch of events

otrosien:
👍

MALPI:
👍

PR #320, opened 2022-04-13
Adds ZStandard compression support

otrosien:
Adds ZStandard compression support, based on the same library that also Nakadi is using (com.github.luben:zstd-jni).

* Added compression support for both GZip and ZStandard
* Limits compression to `POST` requests, as [Nakadi only supports it for those](https://github.com/zalando/nakadi/blob/master/core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java#L52). Luckily Fahrschein does not send any PATCH or PUT anyway. 
* Introduces new Header constants for Content-Encoding and Accept-Encoding.

Closes #310 

## Tasks

* [x] Code implementation & unit tests
* [x] End-to-end test 
* [x] README documentation

PR #322, opened 2022-04-27
JDK11 client: Avoid copying of headers

otrosien:
Introduces a delegate class to avoid copying of all headers.

Also throws 'UnsupportedOperationExceptions' in case of misuse.

### DIFF
--- a/fahrschein-http-jdk11/src/main/java/org/zalando/fahrschein/http/jdk11/JavaNetBufferingRequest.java
+++ b/fahrschein-http-jdk11/src/main/java/org/zalando/fahrschein/http/jdk11/JavaNetBufferingRequest.java
@@ -53,9 +53,12 @@ final class JavaNetBufferingRequest implements Request {
     @Override
     public Headers getHeaders() {
         return new Headers() {
+
+            private static final String HEADER_IMPLEMENTATION_IS_WRITE_ONLY = "Header implementation is write-only";
+
             @Override
             public List<String> get(String headerName) {
-                return List.of();
+                throw new UnsupportedOperationException(HEADER_IMPLEMENTATION_IS_WRITE_ONLY);
             }
 
             @Override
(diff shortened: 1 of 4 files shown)


fahrschein-http-jdk11/src/main/java/org/zalando/fahrschein/http/jdk11/JavaNetBufferingRequest.java
```diff
@@ -73,27 +76,27 @@ public void put(String headerName, String value) {
...
             @Override
             public long getContentLength() {
-                return 0;
+                throw new UnsupportedOperationException(HEADER_IMPLEMENTATION_IS_WRITE_ONLY);
             }
 
             @Override
             public void setContentLength(long contentLength) {
-                // NO-OP.
+                throw new UnsupportedOperationException("Content-Length to be set by underlying framework");
```
otrosien, line 94: could keep this as no-op, if people think it's better..

fahrschein-http-jdk11/src/main/java/org/zalando/fahrschein/http/jdk11/JavaNetResponse.java
```diff
@@ -28,12 +28,12 @@ public String getStatusText() throws IOException {
             case (200): return "OK";
             case (201): return "Created";
             case (202): return "Accepted";
-            case (203): return "Non Authoritative Information";
+            case (203): return "Non-Authoritative Information";
             case (204): return "No Content";
             case (205): return "Reset Content";
             case (206): return "Partial Content";
             case (207): return "Partial Update OK";
-            case (300): return "Mutliple Choices";
```
otrosien, line 36: that's what you get when copying solutions from stack overflow ;)

otrosien:
JavaNetHeadersDelegateTest is more or less a copy of HeadersImplTest - idea: extract base-class with the common test cases in http-test-support.

fahrschein-http-jdk11/src/test/java/org/zalando/fahrschein/http/jdk11/JavaNetHeadersDelegateTest.java
```diff
@@ -0,0 +1,101 @@
...
+    @Test
+    public void shouldReturnEmptySetOfHeadernames() {
+        final Headers headers = new JavaNetHeadersDelegate(requestBuilder().build().headers());
+        assertEquals(emptySet(), headers.headerNames());
+    }
+
+    @Test
+    public void shouldReturnEmptyListForUnknownHeaders() {
+        final Headers headers = new JavaNetHeadersDelegate(requestBuilder().build().headers());
+        assertEquals(emptyList(), headers.get("Content-Type"));
```
othon2000, line 54: Let's use the constants defined in `org.zalando.fahrschein.http.api.Headers`.
I also suggest adding `Content-Encoding` there, and replacing the occurrences (also fine to delegate the replacement in other modules to a different PR, but it's quite a simple task I'd do it here).

The constants can also be used wherever you're testing with upper or lower case.

otrosien: I've started doing this in the other PR: https://github.com/zalando-nakadi/fahrschein/pull/320 - I'll rebase this one once the zstd PR is merged.

fahrschein-http-spring/src/main/java/org/zalando/fahrschein/http/spring/HeadersAdapter.java
```diff
@@ -18,7 +19,7 @@ class HeadersAdapter implements Headers {
 
     @Override
     public List<String> get(String headerName) {
-        return headers.get(headerName);
+        return Optional.ofNullable(headers.get(headerName)).orElse(Collections.emptyList());
```
otrosien, line 22: Found a surprisingly differing behaviour in the get(headerName) implementations: Spring was returning null, but in our basic HeadersImpl (reference impl?) we were specifying that get on an unknown header would return an emptyList.

othon2000 (approved):

otrosien:
👍

PR #323, opened 2022-04-29
Switch from maven to gradle 

otrosien:
Rationale: Our current multi-module build with maven has flaws, and we could benefit from more powerful tooling for our needs. We currently have...

* complex submodule interdependencies, which cause circular dependency problems in maven, because maven doesn't have first-class support of test-fixtures and test-fixture-based project dependencies
* an issue with publishing: Our test fixture projects get published as maven artefacts, as well as fahrschein-parent
* noisy build output
* no differentiation between dependencies that become part of our API and dependencies that are implementation details https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies - leading to issues like https://github.com/zalando-nakadi/fahrschein/issues/328

Tasks:
* [x] Add gradle project setup
* [x] Run "build" github workflow  with gradle 
* [x] Run "publish" github workflow with gradle
* [x] Remove pom.xmls
* [x] Support CVE scanning
* [x] Support test coverage reporting 
* [x] Document common tasks like building, publishing and CVE scanning
* [x] Clean up the PR branch


fahrschein/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'fahrschein.java-conventions'
+    id 'fahrschein.maven-publishing-conventions'
+}
+
+dependencies {
+    api project(':fahrschein-http-api')
+    runtimeOnly project(':fahrschein-http-simple')
```
otrosien, line 8: Make this api https://github.com/zalando-nakadi/fahrschein/issues/328

.github/workflows/cve-scan.yml
```diff
@@ -0,0 +1,18 @@
+name: CVE scan
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
otrosien, line 4: This means we'll check every day and get notified if the cve scan fails.

README.md
```diff
@@ -278,21 +278,61 @@ If you have questions, concerns, bug reports, etc, please file an issue in this
...
 * A local Docker installation for running integration tests
 
-When developing, make sure to run unit and integration tests with `mvn verify`.
+When developing, make sure to run unit and integration tests with `./gradlew test`.
+
+### Understanding the build
+
+We use Gradle [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html) to share common build logic across multiple subprojects. These are [fahrschein.java-conventions.gradle](./buildSrc/src/main/groovy/fahrschein.java-conventions.gradle) for common java properties, and [fahrschein.maven-publishing-conventions.gradle](./buildSrc/src/main/groovy/fahrschein.maven-publishing-conventions.gradle) for subprojects that are released as maven artefacts.
+
+### Bumping dependency versions
```
otrosien, line 292: document compatibility baseline?

build.gradle
```diff
@@ -0,0 +1,14 @@
...
+    id 'com.github.ben-manes.versions' version '0.42.0'
+    id 'org.owasp.dependencycheck' version '7.1.0.1'
+}
+
+
+// CVE vulnerability scanning
+// run: ./gradlew dependencyCheckAggregate
+
+dependencyCheck {
+    failBuildOnCVSS = '9'
```
otrosien, line 13: there was no cvs score baseline set before, shall we start with 9?

buildSrc/src/main/groovy/fahrschein.java-conventions.gradle
```diff
@@ -0,0 +1,60 @@
...
+java.targetCompatibility = JavaVersion.VERSION_1_8
+
+tasks.withType(JavaCompile) {
+    options.compilerArgs << '-parameters'
+    options.compilerArgs << '-Xlint:all'
+}
+
+dependencies {
+    implementation("org.slf4j:slf4j-api:${property('slf4jV')}") {
+        because "we want slf4j-api to be available everywhere"
```
otrosien, line 23: isn't this cute? :)

fahrschein-example/build.gradle
```diff
@@ -9,16 +9,16 @@ dependencies {
     implementation project(':fahrschein-http-spring')
     implementation project(':fahrschein-inmemory')
     implementation project(':fahrschein-typeresolver')
-    implementation "com.squareup.okhttp3:okhttp:${property('okHttpV')}"
-    implementation "org.apache.httpcomponents:httpclient:${property("apacheHttpV")}"
-    implementation "org.springframework:spring-web:${property('springV')}"
+    implementation "com.squareup.okhttp3:okhttp:${property('okhttp.version')}"
+    implementation "org.apache.httpcomponents:httpclient:${property("apachehttp.version")}"
+    implementation "org.springframework:spring-web:${property('spring.version')}"
     implementation 'org.zalando:jackson-datatype-money:0.6.0'
```
otrosien, line 15: this is sooooo outdated!

buildSrc/src/main/groovy/fahrschein.java-conventions.gradle
```diff
@@ -0,0 +1,60 @@
...
+java.targetCompatibility = JavaVersion.VERSION_1_8
+
+tasks.withType(JavaCompile) {
+    options.compilerArgs << '-parameters'
+    options.compilerArgs << '-Xlint:all'
+}
+
+dependencies {
+    implementation("org.slf4j:slf4j-api:${property('slf4j.version')}") {
+        because "we want slf4j-api to be available everywhere"
```
otrosien, line 23: Dependency reasoning... isn't this cute? :)

fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
```diff
@@ -172,26 +174,8 @@ private static void subscriptionListen(ObjectMapper objectMapper, Listener<Sales
...
-        final CloseableHttpClient httpClient = HttpClients.custom()
-                .setDefaultRequestConfig(requestConfig)
-                .setDefaultConnectionConfig(connectionConfig)
-                .setConnectionTimeToLive(30, TimeUnit.SECONDS)
-                .disableAutomaticRetries()
-                .disableRedirectHandling()
-                .setMaxConnTotal(8)
-                .setMaxConnPerRoute(2)
-                .build();
+        final CloseableHttpClient httpClient = HttpClients.createDefault();
```
otrosien, line 178: simplified this because without this code we can even run against some lower apache httpclient version.

otrosien: we should probably add some note about sizing the connection pooling correctly.

.github/workflows/ci.yml
```diff
@@ -15,10 +15,12 @@ jobs:
         with:
           distribution: 'adopt-hotspot'
           java-version: '11'
-          cache: 'maven'
+      - uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 7.4.1
```
cberg-zalando, line 20: Is it possible to manage these versions at a central place to avoid having to update more than one place, e.g. a new Gradle version is released?

.gitignore
```diff
@@ -9,3 +9,7 @@ target/
 .settings/
 .vscode
 bin/
+
+#gradle
```
cberg-zalando, line 13: Is this a debug leftover?

buildSrc/src/main/groovy/fahrschein.maven-publishing-conventions.gradle
```diff
@@ -0,0 +1,85 @@
...
+            versionMapping {
+                usage('java-api') {
+                    fromResolutionOf('runtimeClasspath')
+                }
+                usage('java-runtime') {
+                    fromResolutionResult()
+                }
+            }
+            pom {
+                url = 'http://github.com/zalando-nakadi/fahrschein'
```
cberg-zalando, line 24:
```suggestion
                url = 'https://github.com/zalando-nakadi/fahrschein'
```

buildSrc/src/main/groovy/fahrschein.maven-publishing-conventions.gradle
```diff
@@ -0,0 +1,85 @@
...
+                inceptionYear = '2006'
+                licenses {
+                    license {
+                        name = 'Apache License, Version 2.0'
+                        url = 'http://www.apache.org/licenses/LICENSE-2.0.txt'
```
cberg-zalando, line 29:
```suggestion
                        url = 'https://www.apache.org/licenses/LICENSE-2.0.txt'
```

fahrschein-e2e-test/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+    testImplementation "com.fasterxml.jackson.core:jackson-databind:${property('jackson.version')}"
+    testImplementation "com.fasterxml.jackson.module:jackson-module-parameter-names:${property('jackson.version')}"
+    testImplementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${property('jackson.version')}"
+    testImplementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${property('jackson.version')}"
+}
+
+test {
+    onlyIf { ! project.hasProperty('e2e.skip') }
+    maxParallelForks = 1
+    systemProperty "COMPOSE_PROVIDED", project.hasProperty("e2e.composeProvided") ? "true" : "false"
```
cberg-zalando, line 25: So it is not possible to just store a boolean value in this property?

fahrschein-example/build.gradle
cberg-zalando: You mean the last update of this library was 2015?

fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
cberg-zalando: So maybe just add this code as an example to the README file?

fahrschein-http-jdk11/build.gradle
```diff
@@ -0,0 +1,14 @@
...
+
+dependencies {
+    api project(':fahrschein-http-api')
+    testImplementation(testFixtures(project(':fahrschein-http-test-support')))
+}
+
+java.sourceCompatibility = JavaVersion.VERSION_11
+java.targetCompatibility = JavaVersion.VERSION_11
+
+publishing.publications.maven.pom.description = 'Fahrschein HTTP Client using JDK11\'s HttpClient'
```
cberg-zalando, line 14: Wouldn't it in this case be better to use the `"` quotations to get rid of quoting the `'`?

fahrschein-http-spring/build.gradle
```diff
@@ -0,0 +1,13 @@
...
+}
+
+dependencies {
+    api project(':fahrschein-http-api')
+    api "org.springframework:spring-web:${property('spring.version')}"
+    testImplementation "com.squareup.okhttp3:okhttp:${property('okhttp.version')}"
+    testImplementation(testFixtures(project(':fahrschein-http-test-support')))
+}
+
+publishing.publications.maven.pom.description = 'Fahrschein HTTP Spring Adapter'
```
cberg-zalando, line 13: What about using a common description for the HTTP client projects. So maybe going with `Fahrschein HTTP Client using Spring Adapter`.

fahrschein/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    id 'fahrschein.maven-publishing-conventions'
+}
+
+dependencies {
+    api project(':fahrschein-http-api')
+    api project(':fahrschein-http-simple')
+    api "com.fasterxml.jackson.core:jackson-databind:${property('jackson.version')}"
+    implementation "com.fasterxml.jackson.module:jackson-module-parameter-names:${property('jackson.version')}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${property('jackson.version')}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${property('jackson.version')}"
```
cberg-zalando, line 12: Pulling in the same dependency twice? Should the first one be `-core`?

gradle.properties
```diff
@@ -0,0 +1,12 @@
...
+
+## dependency versions
+jackson.version=2.13.1
+slf4j.version=1.7.25
+spring.version=5.3.18
+mockito.version=4.3.1
+junit.version=4.13.1
+testContainersV=1.16.2
+okhttp.version=4.9.3
+apachehttp.version=4.5.13
```
cberg-zalando, line 12: End of file newline issue.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.1-bin.zip
```
cberg-zalando, line 3: IntelliJ recommends using the `-all` distribution and I personally recommend using the `distributionSha256Sum` parameter to just make sure that you don't pull a manipulated Gradle distribution.

fahrschein-example/build.gradle
otrosien: The version we're pulling in here (0.6.0) is from 2015.  Last release (1.3.0) is from Nov 2021.

cberg-zalando: Is there a blocker why we cannot update it to the latest one?

.github/workflows/ci.yml
otrosien: The alternative is using the wrapper, but this will mean we'd be downloading and extracting the wrapper every time we run the CI. I don't think it's critical to have this in multiple places. If it becomes a chore we can still put a helper script into the repo that updates all occasions.

.gitignore
otrosien: no, not a debug leftover. A comment on "here start the gradle-based gitignore entries"

fahrschein-e2e-test/build.gradle
otrosien: system properties are strings, and this translation is to make it a flag on command-line and save some typing.

fahrschein-http-spring/build.gradle
otrosien: I copied the descriptions from the pom.xml, having a common format is reasonable. 👍

fahrschein-http-jdk11/build.gradle
otrosien: 👍

gradle/wrapper/gradle-wrapper.properties
otrosien: The "all" distribution is 40MB bigger, what benefit do you see in pulling it instead of bin? Regarding distributionSha256Sum, I wasn't aware this feature existed, probably doesn't hurt, but also sounds a bit paranoid.

cberg-zalando: It was just a recommendation by IntelliJ. I think, it was typically something regarding debugging IIRC, but I am not fully sure anymore.

fahrschein-example/build.gradle
otrosien: no, but I wouldn't want to upgrade dependencies as part of this PR.

gradle/wrapper/gradle-wrapper.properties
otrosien: let's start with the bin. Can later be changed in case we find it useful.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,3 +3,4 @@ distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.1-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionSha256Sum=e5444a57cda4a95f90b0c9446a9e1b47d3d7f69057765bfb54bd4f482542d548
```
cberg-zalando, line 6: And another trailing newline issue.

otrosien: what is this problem really about?

cberg-zalando: I just quote from Stackoverflow:

> It indicates that you do not have a newline (usually '\n', aka CR or CRLF) at the end of file.
>
> That is, simply speaking, the last byte (or bytes if you're on Windows) in the file is not a newline.
>
> The message is displayed because otherwise there is no way to tell the difference between a file where there is a newline at the end and one where is not. Diff has to output a newline anyway, or the result would be harder to read or process automatically.
>
> Note that it is a good style to always put the newline as a last character if it is allowed by the file format. Furthermore, for example, for C and C++ header files it is required by the language standard.
>
> from https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

buildSrc/src/main/groovy/fahrschein.maven-publishing-conventions.gradle
```diff
@@ -0,0 +1,85 @@
...
+    def signingPassword = System.getenv("GPG_PASSPHRASE")
+    useInMemoryPgpKeys(signingKey, signingPassword)
+    sign publishing.publications.maven
+}
+
+javadoc {
+    if(JavaVersion.current().isJava9Compatible()) {
+        options.addBooleanOption('html5', true)
+    }
+}
```
cberg-zalando, line 85: Here is still a newline issue.

fahrschein-example/src/main/java/org/zalando/fahrschein/example/Main.java
otrosien: done

.gitignore
otrosien: kind of a section header, as with "#vscode" above.

buildSrc/src/main/groovy/fahrschein.maven-publishing-conventions.gradle
otrosien: fixed.

.gitignore
cberg-zalando: Ah. Then at least for me it would help if you add a space in between the `#` and the header. Then it does not look like the value is commented out.

README.md
otrosien: done.

cberg-zalando:
:+1:

buildSrc/src/main/groovy/fahrschein.java-conventions.gradle
```diff
@@ -0,0 +1,60 @@
...
+repositories {
+    mavenCentral()
+}
+
+group = 'org.zalando'
+version = project.property('project.version') + (project.hasProperty('release') ? '' : '-SNAPSHOT')
+java.sourceCompatibility = JavaVersion.VERSION_1_8
+java.targetCompatibility = JavaVersion.VERSION_1_8
+
+tasks.withType(JavaCompile).configureEach {
```
cberg-zalando, line 16: You have a configuration block for `JavaCompile` here and at the end. You should be able to merge them into one, no?!

fahrschein/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'fahrschein.java-conventions'
+    id 'fahrschein.maven-publishing-conventions'
+}
+
+dependencies {
+    api project(':fahrschein-http-api')
+    api project(':fahrschein-http-simple')
+    api "com.fasterxml.jackson.core:jackson-databind:${property('jackson.version')}"
+    implementation "com.fasterxml.jackson.module:jackson-module-parameter-names:${property('jackson.version')}"
```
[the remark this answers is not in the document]
cberg-zalando: If fahrschein explicitly uses these Jackson modules, I think, there is no other was to add it to the implementation detail, as otherwise the library user would need to make sure that these are available as a runtime dependency.

Jackson is able to load additional modules via `ServiceLoader`, but this would only work if fahrschein would only use the modules when they are there and fallback to some defaults when they are missing.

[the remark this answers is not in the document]
cberg-zalando: Or were you also referring to the core Jackson libraries being used?

fahrschein-e2e-test/build.gradle
```diff
@@ -25,5 +25,6 @@ test {
     systemProperty "COMPOSE_PROVIDED", project.hasProperty("e2e.composeProvided") ? "true" : "false"
 }
 
-java.sourceCompatibility = JavaVersion.VERSION_11
-java.targetCompatibility = JavaVersion.VERSION_11
+tasks.withType(JavaCompile).configureEach {
+    options.release = 11
```
cberg-zalando, line 29: So setting this option is better in making sure that we go with Java 11 then the options above?

otrosien: The other options are incompatible with gradle toolchains somehow, and this is actually the successor for java9+ and the recommended way going forward.

cberg-zalando: Thanks for the explanation. All clear now. :smile:

fahrschein-metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id 'fahrschein.java-conventions'
+    id 'fahrschein.maven-publishing-conventions'
+}
+
+dependencies {
+    api project(':fahrschein')
+    api 'io.dropwizard.metrics:metrics-core:3.1.2'
```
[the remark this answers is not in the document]
cberg-zalando: I also asked about this at another location and @otrosien said that he wants to do this in a separate ticket.

build.gradle
```diff
@@ -0,0 +1,14 @@
+plugins {
+    id 'eclipse'
+    id 'idea'
+    id 'com.github.ben-manes.versions' version '0.42.0'
```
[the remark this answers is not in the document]
cberg-zalando: You mean with dependabot integration?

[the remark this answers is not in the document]
otrosien: There's dependabot, yes, but that works slightly differently (and both solutions aren't mutually exclusive).

build.gradle
```diff
@@ -0,0 +1,14 @@
...
+    id 'org.owasp.dependencycheck' version '7.1.0.1'
```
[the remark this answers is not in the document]
otrosien: CVE score checking was part of our maven setup (although not automated via Github). That's why I also included in the gradle setup.

buildSrc/build.gradle
```diff
@@ -0,0 +1,7 @@
+plugins {
+    id 'groovy-gradle-plugin'
```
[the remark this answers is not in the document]
otrosien: buildSrc is for shaing the common settings in our projects ( see https://docs.gradle.org/current/samples/sample_convention_plugins.html ) This "groovy-gradle-plugin" is the plugin enabling that feature.

fahrschein-e2e-test/src/test/java/org/zalando/fahrschein/e2e/NakadiTestWithDockerCompose.java
```diff
@@ -29,7 +29,7 @@
 
 public abstract class NakadiTestWithDockerCompose {
 
-    private static final boolean COMPOSE_PROVIDED = System.getProperty("COMPOSE_PROVIDED") != null;
+    private static final boolean COMPOSE_PROVIDED = Boolean.parseBoolean(System.getProperty("COMPOSE_PROVIDED"));
```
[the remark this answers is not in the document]
otrosien: yes, we (still) do. But we keep an option to run the docker-compose setup outside of our build, just like we did with maven. Reason for this change, it was a bit more convenient to use parseBoolean with gradle, capturing the empty string case, than to check for null.

build.gradle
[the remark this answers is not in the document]
otrosien: If you want dependabot, let's discuss and open a ticket to integrate it.

MALPI:
👍

cberg-zalando:
:+1:

fahrschein/build.gradle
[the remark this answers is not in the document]
otrosien: background read: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

So, for example 'jackson-module-parameter-names' will be part of the runtime classpath (therefore usable by fahrschein) but projects depending on fahrschein will not be able to use it in their compilation unless explicitly called as direct dependency.

otrosien:
👍

Issue #328, opened 2022-05-10
fahrschein -> fahrschein-http-simple dependency issue


otrosien:
**Describe the bug**

The dependency between fahrschein and fahrschein-http-simple has become a "runtime" dependency (*). This will be interpreted as "implementation detail" by gradle, therefore not allowing direct access to the SimpleRequestFactory class.

(*) https://github.com/zalando-nakadi/fahrschein/blame/master/fahrschein/pom.xml#L22

**To Reproduce**
1. Create a minimal gradle project (error reported on gradle 7.1.1).
2. Add fahrschein dependency ` dependencies { implementation 'org.zalando:fahrschein:0.21.0' } `
3. Try to construct the NakadiClient by using the SimpleRequestFactory.

**Expected behavior**

The goal of the fahrschein-http-simple dependency is to provide a one-stop-shop experience when including fahrschein into the project.

**Additional context**

For consideration: Evaluate the breaking change of excluding the fahrschein-http-simple dependency altogether, to encourage teams to switch to the fahrschein-http-jdk11 implementation.

**Workaround**

For users, the easiest way to work-around that issue, is to add their desired http-client implementation as direct dependency, e.g. if a project is on JDK11+, potentially switch to the new JDK11 implementation.

PR #329, opened 2022-05-10
Reverts the fahrschein-http-simple to runtime dependency

otrosien:
Although technically correct (there is no dependency at build time),
we want to expose the SimpleRequestFactory to users at their compile-time (!), which is
prohibited by gradle's interpretation of the pom.xml, if a dependency
is marked as runtime.

https://github.com/zalando-nakadi/fahrschein/issues/328

In a closer look, this is a valid interpretation of [the spec](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope):

**runtime** This scope indicates that the dependency is not required for compilation, but is for execution. Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath.

### DIFF
--- a/fahrschein/pom.xml
+++ b/fahrschein/pom.xml
@@ -19,7 +19,6 @@
             <groupId>org.zalando</groupId>
             <artifactId>fahrschein-http-simple</artifactId>
             <version>${project.version}</version>
-            <scope>runtime</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


MALPI:
👍

otrosien:
👍

PR #345, opened 2022-07-13
Remove outdated documentation to handle errors

anilbharadia:
Remove the outdated documentation in the readme file, which is about how to handle errors using `withErrorHandler()` which is removed in PR #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,17 +168,6 @@ future.cancel(true);
 final Future<?> future2 = executorService.submit(runnable);
 ```
 
-### Handling data binding problems
-
-You might want to ignore events that could not be mapped to your domain objects by Jackson, instead of having these events block all further processing.
-To achieve this you can implement the `onMappingException` method of the `ErrorHandler` interface handle the `JsonMappingException` yourself.
-
-```java
-nakadiClient.stream(eventName)
-        .withErrorHandler(e -> {...})
-        .listen(SalesOrderPlaced.class, listener);
-```
-
 ## `RequestFactory` implementations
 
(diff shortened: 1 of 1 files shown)


otrosien:
@anilbharadia thanks for the contribution. I would assume we need to document the new way of handling mapping exceptions instead.

MALPI:
Yeah Agree here, but I have to admit that the explicit error handling with a dedicated `ErrorHandler` was more elegant. This would encourage teams to make use of a DLQ if required for example. At the same time they could still do it now....

MALPI:
Would it make sense to adjust the documentation in a way that encourages clients to do exception handling accordingly and add a link to some kind of best practices or so?

otrosien:
From the docs:

> ## Exception handling
> 
> Exception handling while streaming events follows some simple rules
> 
>  - `IOException` and its subclasses are treated as temporary failures. Library will automatically recover from this 
>  type of failure by retrying processing attempt after time interval specified by the `BackoffStrategy`.
>  **Special case**: if an `IOException` happens during opening the initial connection, this is treated as configuration 
>  problem (wrong host name or missing scopes). In this case processing will be aborted and exception will be re-thrown. 
>  - If `listener` throws `RuntimeException` streaming of events will be aborted. User is responsible for handling these exceptions.
>  Library code itself will not throw `RuntimeException`s.
>  - Exceptions in other client methods are not automatically retried

MALPI:
My bad, only saw the part that we dropped.

MALPI:
👍

otrosien:
👍

PR #348, opened 2022-07-15
Architectural unit tests

otrosien:
A library project like Fahrschein should ensure that over time no architectural antipattern is introduced. This includes unwanted exposure of APIs, or other design flaws. For details see https://www.archunit.org/

As part of that we should review:

* which classes should be public, which methods should be public?
* which of the public classes do we want to mark as final?
* are we returning implementations instead of interfaces anywhere?
* what are expected or unexpected class dependencies (don't think there are, but could codify the existing deps on package level for example)


otrosien:
interesting test, but not really useful for us.

PR #349, opened 2022-07-15
Check code against "Forbidden APIs"

otrosien:
Some Java APIs are considered bad practice because they rely on default locale, charset or timezone. This PR fixes the code to not use these APIs and ensures that no forbidden API gets introduced going forward.

For details, see https://github.com/policeman-tools/forbidden-apis/wiki

I want to highlight that, even though the changes caused by introducing these checks at the moment look like false-positives, it may save us making mistakes going forward.

### DIFF
--- a/fahrschein-http-apache/src/main/java/org/zalando/fahrschein/http/apache/HttpComponentsRequestFactory.java
+++ b/fahrschein-http-apache/src/main/java/org/zalando/fahrschein/http/apache/HttpComponentsRequestFactory.java
@@ -33,7 +33,7 @@ import java.net.URI;
  * @author Juergen Hoeller
  * @author Joern Horstmann
  */
-public final class HttpComponentsRequestFactory implements RequestFactory {
+public class HttpComponentsRequestFactory implements RequestFactory {
 
     private final HttpClient httpClient;
     private final ContentEncoding contentEncoding;
--- a/fahrschein-http-apache/src/main/java/org/zalando/fahrschein/http/apache/HttpComponentsRequestFactory.java
+++ b/fahrschein-http-apache/src/main/java/org/zalando/fahrschein/http/apache/HttpComponentsRequestFactory.java
@@ -33,7 +33,7 @@ import java.net.URI;
  * @author Juergen Hoeller
  * @author Joern Horstmann
  */
-public class HttpComponentsRequestFactory implements RequestFactory {
+public final class HttpComponentsRequestFactory implements RequestFactory {
 
     private final HttpClient httpClient;
     private final ContentEncoding contentEncoding;
--- a/fahrschein-http-simple/src/main/java/org/zalando/fahrschein/http/simple/SimpleRequestFactory.java
+++ b/fahrschein-http-simple/src/main/java/org/zalando/fahrschein/http/simple/SimpleRequestFactory.java
@@ -21,7 +21,7 @@ import java.net.URLConnection;
  * @author Joern Horstmann
  * @see java.net.HttpURLConnection
  */
-public final class SimpleRequestFactory implements RequestFactory {
+public class SimpleRequestFactory implements RequestFactory {
 
     private static final int DEFAULT_CONNECT_TIMEOUT = 500;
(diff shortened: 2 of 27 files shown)


fahrschein-http-test-support/src/testFixtures/java/org/zalando/fahrschein/http/test/AbstractRequestFactoryTest.java
```diff
@@ -24,6 +24,7 @@
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
```
otrosien, line 27:
```suggestion
```

fahrschein-http-apache/src/main/java/org/zalando/fahrschein/http/apache/HttpComponentsRequestFactory.java
```diff
@@ -33,7 +33,7 @@
  * @author Juergen Hoeller
  * @author Joern Horstmann
  */
-public class HttpComponentsRequestFactory implements RequestFactory {
+public final class HttpComponentsRequestFactory implements RequestFactory {
```
otrosien, line 36:
```suggestion
public class HttpComponentsRequestFactory implements RequestFactory {
```

otrosien: revert unrelated change.

fahrschein-http-simple/src/main/java/org/zalando/fahrschein/http/simple/SimpleRequestFactory.java
```diff
@@ -21,7 +21,7 @@
  * @author Joern Horstmann
  * @see java.net.HttpURLConnection
  */
-public class SimpleRequestFactory implements RequestFactory {
+public final class SimpleRequestFactory implements RequestFactory {
```
otrosien, line 24:
```suggestion
public class SimpleRequestFactory implements RequestFactory {
```

fahrschein/src/main/java/org/zalando/fahrschein/IdentityAcceptEncodingRequestFactory.java
```diff
@@ -7,7 +7,7 @@
 import java.io.IOException;
 import java.net.URI;
 
-public class IdentityAcceptEncodingRequestFactory implements RequestFactory {
+public final class IdentityAcceptEncodingRequestFactory implements RequestFactory {
```
otrosien, line 10:
```suggestion
public class IdentityAcceptEncodingRequestFactory implements RequestFactory {
```

otrosien:
👍

fahrschein/src/main/java/org/zalando/fahrschein/ManagedCursorManager.java
```diff
@@ -143,7 +144,7 @@ public void onSuccess(String eventName, Cursor cursor) throws IOException, Curso
     }
 
     private String inputStreamToString(InputStream inputStream) {
-        return new BufferedReader(new InputStreamReader(inputStream))
+        return new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))
```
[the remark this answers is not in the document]
otrosien: a) this method is called in case of errors, so maybe not critical that we default to UTF-8 here instead of the actual charset used b) the charset in the content-type response header may or may not be present for indicating the encoding. It's not something we can easily introspect here, at least as of now.

What do you mean with the nio import should not be used?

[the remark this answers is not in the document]
otrosien: one thing is for sure: using the process default encoding (what the code was doing before) is probably worse than fixing this to UTF-8 at this line of code.

MALPI:
👍

otrosien:
👍

PR #361, opened 2022-09-08
Drain InputStream on closing

### LINKED ISSUE #303: Drain InputStream on closing
The HTTP client facade code is original from spring, and the code for closing has diverged since. Check if our code needs an update. It's potentially a bug fix around resource leaks.

https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein-http-simple/src/main/java/org/zalando/fahrschein/http/simple/SimpleResponse.java#L73

https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/client/SimpleClientHttpResponse.java#L94

otrosien:
The HTTP client facade code is original from spring, and the code for closing has diverged since. They have introduced a bug-fix to ensure that the InputStreams are properly drained, so the connections can be given back to the connection pools without resource leaks.

Closes #303

Changes:
* New StreamUtils helper class with static method to "drain"
* Calling drain() in fahrschein-http-simple and fahrschein-http-jdk11
* Slight refactoring of the apache client, to capture its InputStream for draining.


otrosien:
.. not sure if it's possible to create test-cases for this. 🤷

MALPI:
👍

otrosien:
👍

PR #364, opened 2022-09-15
Revamp Fahrschein examples

otrosien:
Strips down the examples to one consumer and one producer test.


otrosien:
A few observations from rewriting the examples: For consuming, I like the wrapper. It feels very intuitive and non-intrusive to use, although it makes exception handling a bit awkward (one needs to throw an IOException to let Fahrschein retry, which is not possible via lambdas)

For the producing part, it still feels quite tedious to extract the span context into metadata...

MALPI (commented):
I didn't check in detail if all the use cases that we have in the "old" example Main are still covered in the new one. I agree though that it looks way more organized though and therefore think it's fine to go ahead with this.

MALPI:
👍

otrosien:
👍

PR #365, opened 2022-09-16
Fix forbiddenApi usage of DateTime.now in fahrschein-example

otrosien:
Fix forbiddenApi usage of DateTime.now in fahrschein-example. Missed because I didn't rebase the forbiddenApis change before merging.


otrosien:
👍

MALPI:
👍

PR #367, opened 2022-10-11
Add instrumentation to publishing

otrosien:
Work-in-progress: Based on semantic conventions we should be creating "producer"-tagged spans when publishing events. At the moment, we don't support our API clients in any way for doing this.


fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedNakadiPublisher.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final NakadiPublisher client;
+    private final Tracer tracer;
+
+    public InstrumentedNakadiPublisher(NakadiPublisher client, Tracer tracer) {
+        this.client = client;
+        this.tracer = tracer;
+    }
+
+    public <T> void publish(String eventName, List<T> events, Span parentSpan) throws IOException {
```
otrosien, line 27: Up to debate: apparently opentelemetry prefers inferring the parent span from the "activeSpan" context. From my understanding this only works in a thread-scoped execution flow, so I would rather prefer explicitly passing the parent here.

fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedNakadiPublisher.java
```diff
@@ -0,0 +1,56 @@
...
+                .setAttribute("messaging.destination_kind", "topic")
+                .setAttribute("messaging.destination", eventName)
+                .setAttribute("messaging.system", "Nakadi")
+                .setAttribute("messaging.message_payload_size", sizeBucket(events.size()))
+                .startSpan();
+        try {
+            client.publish(eventName, events);
+        } catch (Throwable t) {
+            childSpan.setStatus(StatusCode.ERROR);
+            childSpan.recordException(t);
```
otrosien, line 41: Open question: add equivalent exception handling in the opentracing impl? One could argue that we're re-throwing the exception and expecting client logic to record exception details...

fahrschein/src/main/java/org/zalando/fahrschein/NakadiPublisher.java
```diff
@@ -0,0 +1,8 @@
+package org.zalando.fahrschein;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface NakadiPublisher {
```
otrosien, line 6: not really necessary in this context, but I found it actually quite neat to define an interface for all the main entrypoints of NakadiClient.

otrosien: We should use the interface instead of the actual class wherever possible.

fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedNakadiPublisher.java
```diff
@@ -0,0 +1,54 @@
...
+            throw t;
+        } finally {
+            childSpan.end();
+        }
+    }
+
+    // @VisibleForTesting
+    // changes must be applied to both OpenTracing and OpenTelemetry implementations.
+    static String sizeBucket(int size) {
```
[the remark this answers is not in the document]
otrosien: added a comment for the bucketing. It's based on our semantic conventions for messaging systems.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiPublisher.java
otrosien: agreed out-of-band: we'll introduce these interfaces.

fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedNakadiPublisher.java
otrosien: @MALPI I've been reading a bit more about OpenTelemetry. They are really trying hard to be hidden from the actual business APIs, I think it makes sense to follow this convention. If people are running in a task-based instead of thread-based environment, they'll need to create a parent span around this call in order to preserve the context (or find a different way to set the active parent span).

MALPI:
👍

fahrschein/src/main/java/org/zalando/fahrschein/SubscriptionBuilder.java
```diff
@@ -13,7 +13,8 @@
 import static org.zalando.fahrschein.Preconditions.checkArgument;
 import static org.zalando.fahrschein.Preconditions.checkState;
 
-public class SubscriptionBuilder {
+
+public final class SubscriptionBuilder {
```
otrosien, line 17:
```suggestion
public class SubscriptionBuilder {
```

fahrschein/src/main/java/org/zalando/fahrschein/SubscriptionBuilder.java
```diff
@@ -26,11 +26,11 @@ public class SubscriptionBuilder {
     @Nullable
     private final Authorization authorization;
 
-    SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames) {
+    public SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames) {
         this(nakadiClient, applicationName, eventNames, DEFAULT_CONSUMER_GROUP, SubscriptionRequest.Position.END, null, null);
     }
 
-    private SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, @Nullable List<Cursor> initialCursors, @Nullable Authorization authorization) {
+    public SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, @Nullable List<Cursor> initialCursors, @Nullable Authorization authorization) {
```
otrosien, line 33:
```suggestion
    private SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames, String consumerGroup, SubscriptionRequest.Position readFrom, @Nullable List<Cursor> initialCursors, @Nullable Authorization authorization) {
```

fahrschein/src/main/java/org/zalando/fahrschein/SubscriptionBuilder.java
```diff
@@ -26,11 +26,11 @@ public class SubscriptionBuilder {
     @Nullable
     private final Authorization authorization;
 
-    SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames) {
+    public SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames) {
```
otrosien, line 29:
```suggestion
   SubscriptionBuilder(NakadiClient nakadiClient, String applicationName, Set<String> eventNames) {
```

fahrschein/src/main/java/org/zalando/fahrschein/SubscriptionBuilder.java
```diff
@@ -67,3 +67,4 @@ public Subscription subscribe() throws IOException {
         return nakadiClient.subscribe(applicationName, eventNames, consumerGroup, readFrom, initialCursors, authorization);
     }
 }
+
```
otrosien, line 70:
```suggestion
```

otrosien:
I'm moving this back to draft state. Maybe we want to offer annotations for this instead...?

PR #370, opened 2022-11-03
Removing Othon Crelier from the maintainers' list

othon2000:
Othon will no longer be a maintainer of Fahrschein after switching back to the management track at Zalando.


MALPI:
👍

otrosien:
👍

MALPI:
👍

PR #372, opened 2022-12-07
Dependency updates

otrosien:
## Changes

* Updates to the latest stable version of all dependencies including gradle and gradle-plugins
* Exceptions: slf4j (major bump), spring (major bump) which we should spare for our next major version upgrade
* Fixes some inconsistencies in the build gradle definitions, removing some duplicate version definitions and unused version properties
* Unrelated: Fixes the port exposure in docker-compose.yaml to string, as podman-compose is more picky about this being a string and not a number

Does not fix the CVE failure reported in the nightly scan: https://github.com/zalando-nakadi/fahrschein/actions/runs/3634792823/jobs/6133259207 - I looked at it and it seems like this is a false positive.


otrosien:
👍

MALPI:
👍

otrosien:
👍

PR #373, opened 2022-12-07
To log "info" in NakadiReader

otrosien:
In every Nakadi batch, an `info.debug` field might be available which helps to understand what's happening to the connection. (e.g. if connection is going to be closed from server side)

In this small PR I'm adding this capability to the `NakadiReader` to log value of `info.debug` instead of ignoring it.

Reopened external PR https://github.com/zalando-nakadi/fahrschein/pull/369 from @sameei


otrosien:
@sameei can you give a bit of context why this feature is important? Is it because Nakadi is closing the connection regularly, and you want to correlate connection-closed error messages with this? Or is it purely educational info for debugging?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiReader.java
```diff
@@ -52,16 +52,55 @@
...
+            BackoffStrategy backoffStrategy,
+            CursorManager cursorManager,
+            Set<String> eventNames,
+            Optional<Subscription> subscription,
+            Optional<Lock> lock,
+            EventReader<T> eventReader,
+            Listener<T> listener,
+            BatchHandler batchHandler,
+            final MetricsCollector metricsCollector,
+            StreamInfoReader streamInfoReader
```
otrosien, line 102: A bit overkill to me - why would someone want to plug their own StreamInfoReader? I think this is only introduced because of testing. Can we find a better way of doing it?

otrosien:
👍

sameei:
Apologies for my inactivity for so long.
Looks good. Thanks for taking care of this, and I will follow the guide you gave about handling new features/fixes/etc. @otrosien

sameei:
👍

MALPI:
👍

PR #377, opened 2023-02-16
Support Spring Framework 6

otrosien:
Fahrschein has two optional dependencies on spring: one (of course) in the fahrschein-http-spring module, and one in fahrschein-test-support where we use a `@Transactional` annotation (from the days when we had jdbc support for cursor handling).

This PR sets Spring as provided dependency, and doesn't change the compile-time dependency but runs integration tests with 4.3.0 - 6.x.

## Background and References

* https://spring.io/blog/2022/11/16/spring-framework-6-0-goes-ga

## Changes

- [x] Change spring-web to become a provided dependency
- [x] Upgrade github actions to test 6.x
- [x] Update documentation


fahrschein-http-spring/build.gradle
```diff
@@ -10,4 +10,8 @@ dependencies {
     testImplementation(testFixtures(project(':fahrschein-http-test-support')))
 }
 
+tasks.withType(JavaCompile).configureEach {
+    options.release = 17
```
bocytko, line 14: There is also [fahrschein-http-jdk11](https://github.com/zalando-nakadi/fahrschein/tree/main/fahrschein-http-jdk11). What will happen to this module?

bocytko:
> Do we want to keep supporting Spring 5? From my understanding, clients can force the spring framework dependency to 5 and fahrschein will still work, but we'll be forcing clients to upgrade to JDK17.

I'd expect security upgrades to be released with JDK11 compatibility. How about releasing 0.30 that would have JDK17 requirement and leave room for 0.24.x for security updates?

fahrschein-http-spring/build.gradle
otrosien: That module is independent of spring and already now has a baseline of JDK11. So, for projects that don't depend on fahrschein-http-spring nothing changes.

otrosien:
> I'd expect security upgrades to be released with JDK11 compatibility. How about releasing 0.30 that would have JDK17 requirement and leave room for 0.24.x for security updates?

The interface we're using in fahrschein didn't even change, so officially we don't even have to upgrade to stay compatible with spring6. We'll discuss details with the maintainer group later.

otrosien:
@bocytko I updated the PR to make Spring a provided dependency, which means any version of Spring will work for us, and we're integration-testing 4.3.0 - 6.x.

.github/workflows/ci.yml
```diff
@@ -30,5 +30,5 @@ jobs:
           flags: unittests
       - name: Test oldest supported library versions
         run: gradle build -Pjackson.version=2.8.0 -Pspring.version=4.3.0.RELEASE -Pokhttp.version=3.3.0 -Papachehttp.version=4.3
-      - name: Test with latest library (micro) updates
```
MALPI, line 33: wouldn't we need another build for spring 5 then? Maybe we drop the spring 4 build?

otrosien: no,  actually we're compatible with all versions of spring with this one build.

otrosien: (and spring 5 is already integration tested in the gradle build run above)

MALPI:
:+1:

otrosien:
Tested with some dummy code via mavenLocal 👍

otrosien:
👍

PR #378, opened 2023-02-17
Add publishing instrumentation

MALPI:
**Problem to solve**

We want to offer means to instrument publishing calls with tracing.

**Considered Approaches**

Instead of having separate Publishers I thought we could introduce some kind of interceptor. The concrete implementation will the go into the telemetry/tracing modules. see #367 

The second thought I had was to extend the `RequestFactory` but there it doesn't work anymore as we don't have context about the execution of the request.

Maybe it would be better to have an explicit `RequestHandler` interface that we can invoke `before` and `after` execution. Like this we could as well easier integrate all the metrics features and last but not least the instrumentation for consumption?

**Implemented Approach**

We introduced the `EventPublishingHandler` Interface that will be invoked before, after and in case of exceptions by the `NakadiClient`.

Like that clients can actually build their own business logic based on the interface to build features. Moreover clients have the option to pass their handler implementation as part of the builder.

**Background and References**
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-name
https://observability.docs.zalando.net/observability-manual/message-queue-clients/#tracing (internal)

### DIFF
--- a/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
+++ b/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
@@ -3,14 +3,12 @@ package org.zalando.fahrschein.http.api;
 import java.util.List;
 
 /**
- * Interface that is offering a way to add additional features around the {@code NakadiClient}.
- *
+ * Interface that offers a way to add additional features around the {@code NakadiClient}.
  * As the methods of this class are invoked on every publish request, we shouldn't
- *
  * - do heavy computations as it might affect performance significantly
  * - do modification of requests
  */
-public interface RequestHandler {
+public interface EventPublishingHandler {
 
     /**
      * This method is called by the {@code NakadiClient} when a request towards Nakadi is going to be sent.
@@ -26,7 +24,7 @@ public interface RequestHandler {
      * This method is invoked after the publishing of events has happened.
      * It can be used to evaluate the given response or others like closing a trace.
      *
-     * @param response
+     * @param response that was received after publishing
      */
     void afterPublish(Response response);
 
--- a/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
+++ b/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
@@ -22,11 +22,9 @@ public interface EventPublishingHandler {
 
     /**
      * This method is invoked after the publishing of events has happened.
-     * It can be used to evaluate the given response or others like closing a trace.
      *
-     * @param response that was received after publishing
      */
-    void afterPublish(Response response);
+    void afterPublish();
 
     /**
      * Invoked when publishing of events failed.
--- a/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
+++ b/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
@@ -1,10 +1,31 @@
 package org.zalando.fahrschein.http.api;
 
+/**
+ * Interface that is offering a way to add additional features around the {@code NakadiClient}
+ */
 public interface RequestHandler {
 
+    /**
+     * This method is called by the {@code NakadiClient} before
+     * a request is going to be executed.
+     * @param request that is going to be executed after the invocation
+     */
     void beforeExecute(Request request);
 
-    void afterExecute(Request request);
+    /**
+     * This method is called by the {@code NakadiClient} after
+     * a request got executed.
+     * @param request that got send to Nakadi
+     * @param response that we received from Nakadi
+     */
+    void afterExecute(Request request, Response response);
 
-    void onError(Request request);
+    /**
+     * This method is called by the {@code NakadiClient} in case
+     * an exception got thrown while sending the request.
+     * The exception itself is rethrown by the client.
+     * @param request that should be sent to Nakadi
+     * @param t Exception that got thrown
+     */
+    void onError(Request request, Throwable t);
 }
--- a/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
+++ b/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
@@ -1,0 +1,10 @@
+package org.zalando.fahrschein.http.api;
+
+public interface RequestHandler {
+
+    void beforeExecute(Request request);
+
+    void afterExecute(Request request);
+
+    void onError(Request request);
+}
--- a/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
+++ b/fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
@@ -1,31 +1,42 @@
 package org.zalando.fahrschein.http.api;
 
+import java.util.List;
+
 /**
- * Interface that is offering a way to add additional features around the {@code NakadiClient}
+ * Interface that is offering a way to add additional features around the {@code NakadiClient}.
+ *
+ * As the methods of this class are invoked on every publish request, we shouldn't
+ *
+ * - do heavy computations as it might affect performance significantly
+ * - do modification of requests
  */
 public interface RequestHandler {
 
     /**
-     * This method is called by the {@code NakadiClient} before
-     * a request is going to be executed.
-     * @param request that is going to be executed after the invocation
+     * This method is called by the {@code NakadiClient} when a request towards Nakadi is going to be sent.
+     * For example, it can be used to record things before that.
+     *
+     * @param eventName that is used for the published events
+     * @param events that are published
+     * @param <T> type of events that we publish
      */
-    void beforeExecute(Request request);
+    <T> void onPublish(String eventName, List<T> events);
 
     /**
-     * This method is called by the {@code NakadiClient} after
-     * a request got executed.
-     * @param request that got send to Nakadi
-     * @param response that we received from Nakadi
+     * This method is invoked after the publishing of events has happened.
+     * It can be used to evaluate the given response or others like closing a trace.
+     *
+     * @param response
      */
-    void afterExecute(Request request, Response response);
+    void afterPublish(Response response);
 
     /**
-     * This method is called by the {@code NakadiClient} in case
-     * an exception got thrown while sending the request.
-     * The exception itself is rethrown by the client.
-     * @param request that should be sent to Nakadi
-     * @param t Exception that got thrown
+     * Invoked when publishing of events failed.
+     *
+     * @param events that failed to be published
+     * @param t the throwable we experienced while publishing
+     * @param <T> type of events that we publish
      */
-    void onError(Request request, Throwable t);
+    <T> void onError(List<T> events, Throwable t);
+
 }
--- a/fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
+++ b/fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
@@ -13,7 +13,7 @@ import java.util.Locale;
 import static io.opentelemetry.api.trace.SpanKind.PRODUCER;
 
 /**
- * Instruments publishing requests to Nakadi.
+ * Instruments publishing requests to Nakadi by making use of OpenTelemetry.
  */
 public class InstrumentedPublishingHandler implements RequestHandler {
 
--- a/fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
+++ b/fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
@@ -4,7 +4,7 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
-import org.zalando.fahrschein.http.api.RequestHandler;
+import org.zalando.fahrschein.http.api.EventPublishingHandler;
 import org.zalando.fahrschein.http.api.Response;
 
 import java.util.List;
@@ -15,7 +15,7 @@ import static io.opentelemetry.api.trace.SpanKind.PRODUCER;
 /**
  * Instruments publishing requests to Nakadi by making use of OpenTelemetry.
  */
-public class InstrumentedPublishingHandler implements RequestHandler {
+public class InstrumentedPublishingHandler implements EventPublishingHandler {
 
     private final Tracer tracer;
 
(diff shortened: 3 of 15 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/NakadiClientBuilder.java
```diff
@@ -48,6 +67,10 @@ public NakadiClientBuilder withCursorManager(CursorManager cursorManager) {
         return new NakadiClientBuilder(baseUri, objectMapper, authorizationProvider, clientHttpRequestFactory, cursorManager);
     }
 
+    public NakadiClientBuilder withTracingInterceptor(List<RequestHandler> requestHandlers) {
```
otrosien, line 70:
```suggestion
    public NakadiClientBuilder withRequestHandlers(List<RequestHandler> requestHandlers) {
```

 but I would actually prefer `addRequestHandler(RequestHandler requestHandler)` and iteratively regstering them.

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
```diff
@@ -0,0 +1,10 @@
+package org.zalando.fahrschein.http.api;
+
+public interface RequestHandler {
+
+    void beforeExecute(Request request);
```
otrosien, line 5:
```suggestion
    void beforeExecution(Request request);
```

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
```diff
@@ -0,0 +1,10 @@
...
+
+    void afterExecute(Request request);
```
otrosien, line 7:
```suggestion
    void afterExecution(Request request);
```

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
```diff
@@ -0,0 +1,10 @@
+package org.zalando.fahrschein.http.api;
+
+public interface RequestHandler {
```
otrosien, line 3: can you add some api documentation what this should be used for?

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
RameshMalla: Maybe just `after`, `before` and `onError` should be enough.

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/RequestHandler.java
RameshMalla: Maybe the method should also take the `response` object. ?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -95,8 +109,13 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
             objectMapper.writeValue(body, events);
         }
 
+        requestHandlers.forEach(requestHandler -> requestHandler.beforeExecute(request));
```
johannesknauft, line 112: I would suggest writing a test for the execution of the handlers.

rob-h-w: Plus one for a test. It should cover the behaviour if `beforeExecute` etc. throws.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClientBuilder.java
rob-h-w: > but I would actually prefer addRequestHandler(RequestHandler requestHandler) and iteratively regstering them.

Why not both? Having a fluent builder is nice for times where the user thinks they have all the handlers at construction time. I'd much rather construct with `builder.withRequestHandlers(myHandlers).build()` than construct-then-modify using `addRequestHandler` when I already have `myHandlers` defined.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -95,8 +109,13 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
             objectMapper.writeValue(body, events);
         }
 
+        eventPublishingHandlers.forEach(requestHandler -> requestHandler.onPublish(eventName, events));
```
RameshMalla, line 112: should we enclose this line within a `try-catch` block to guarantee that  `onError` and `afterPublish` gets executed. ?

fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import io.opentracing.tag.Tags;
+import org.zalando.fahrschein.http.api.EventPublishingHandler;
+import org.zalando.fahrschein.http.api.Response;
+
+import java.util.List;
+import java.util.Locale;
+
+import static io.opentracing.tag.Tags.ERROR;
+
+public class InstrumentedPublishingHandler implements EventPublishingHandler {
```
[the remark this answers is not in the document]
rob-h-w: Yes. The implementation in `...opentelemetry.InstrumentedPublishingHandler` is slightly different because it uses the opentelemetry API instead of the opentracing one.

`sizeBucket` seems to be a common feature of both. <Nit>It could go in the interface as a static or instance method</Nit>.

rob-h-w (approved):

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -109,8 +109,8 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
             objectMapper.writeValue(body, events);
         }
 
-        eventPublishingHandlers.forEach(requestHandler -> requestHandler.onPublish(eventName, events));
         try (final Response response = request.execute()) {
+            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onPublish(eventName, events));
```
RameshMalla, line 113: Now the `onPublish ` handler is executed after the request is executed ` try (final Response response = request.execute())`. is that ok ?

fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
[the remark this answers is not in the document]
MALPI: As the `sizeBucket` calculation is specific to the tracing logic I don't see that it should move to the API module. At the same time, I think it's not worth the overhead to introduce a shared module for that.

MALPI:
👍

RameshMalla:
👍

johannesknauft:
:+1:

RameshMalla (approved):

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.List;
+
+/**
+ * Interface that offers a way to add additional features around the {@code NakadiClient}.
+ * As the methods of this class are invoked on every publish request, we shouldn't
+ * - do heavy computations as it might affect performance significantly
+ * - do modification of requests
+ */
+public interface EventPublishingHandler {
```
otrosien, line 11: can this interface move to the fahrschein module?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
rob-h-w: > Now the `onPublish` handler is executed after the request is executed

Do you mean before? I read lines 114-117 to mean the sequence `onPublish`, `execute` (i.e. execute the request), `afterPublish`.

RameshMalla: It is fixed as part of this commit https://github.com/zalando-nakadi/fahrschein/pull/378/commits/c086a547d94095c8cae2ab48cee51a3c80d19ed3 👍

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param events that are published
+     * @param <T> type of events that we publish
+     */
+    <T> void onPublish(String eventName, List<T> events);
+
+    /**
+     * This method is invoked after the publishing of events has happened.
+     *
+     */
+    void afterPublish();
```
otrosien, line 27: please document the guarantee on this event (does afterPublish always get called, regardless of error?)

fahrschein-http-api/src/main/java/org/zalando/fahrschein/http/api/EventPublishingHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Invoked when publishing of events failed.
+     *
+     * @param events that failed to be published
+     * @param t the throwable we experienced while publishing
+     * @param <T> type of events that we publish
+     */
+    <T> void onError(List<T> events, Throwable t);
```
otrosien, line 36: IIRC there could be partial failures, does this list of events only contain the failed ones?

otrosien: if not, how could a callback for a partial failure look like?

MALPI:
👍

otrosien:
👍

fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingHandler.java
```diff
@@ -21,15 +21,17 @@ public interface EventPublishingHandler {
     <T> void onPublish(String eventName, List<T> events);
 
     /**
-     * This method is invoked after the publishing of events has happened.
+     * This method is invoked after the publishing of events has happened,
+     * but only if the publishing didn't face an Exception otherwise
+     * onError will be invoked.
```
otrosien, line 26: @MALPI then call it "onSuccess()"?

fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
```diff
@@ -0,0 +1,56 @@
...
+                .startSpan();
+        span.makeCurrent();
+    }
+
+    public void onSuccess() {
+        Span.current().end();
+    }
+
+    @Override
+    public <T> void onError(List<T> successEvents, List<T> failedEvents, Throwable t) {
```
rob-h-w, line 44: I thought we were going to go with a heterogeneous list of event containers denoting success/failure, thereby preserving ordering. This was suggested by @otrosien. I think it's a lovely idea.

otrosien: yep. Let's say you have in one batch two events for the same compaction key, the first one fails the second one succeeds, you shouldn't retry the second one. Given two lists it will be impossible to reconstruct the ordering.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -95,8 +108,19 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
             objectMapper.writeValue(body, events);
         }
 
-        try (final Response response = request.execute()) {
+        Response response = null;
+        try {
+            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onPublish(eventName, events));
+            response = request.execute();
             LOG.debug("Successfully published [{}] events for [{}]", events.size(), eventName);
+            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onSuccess());
```
rob-h-w, line 116:
```suggestion
            reversedEventPublishingHandlers.forEach(requestHandler -> requestHandler.onSuccess());
```

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -95,8 +108,19 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
...
+        } catch (Throwable t) {
+            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onError(Collections.EMPTY_LIST, events, t));
```
rob-h-w, line 118:
```suggestion
            reversedEventPublishingHandlers.forEach(requestHandler -> requestHandler.onError(Collections.EMPTY_LIST, events, t));
```

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -59,8 +61,19 @@ public static NakadiClientBuilder builder(URI baseUri, RequestFactory requestFac
         this.objectMapper = objectMapper;
         this.internalObjectMapper = DefaultObjectMapper.INSTANCE;
         this.cursorManager = cursorManager;
+        this.eventPublishingHandlers = new ArrayList<>();
```
rob-h-w, line 64: If we persist a reversed version of `eventPublishingHandlers`, we can do callbacks after the request in the reverse order, preserving visibility of the effects of each handler.

fahrschein/src/main/java/org/zalando/fahrschein/EventPublishingHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Invoked when publishing of events failed (partially).
+     *
+     * @param <T>           type of events that we publish
+     * @param successEvents List of events we successfully published
+     * @param failedEvents  List of events we failed to publish.
+     * @param t             the throwable we experienced while publishing
+     */
+    <T> void onError(List<T> successEvents, List<T> failedEvents, Throwable t);
```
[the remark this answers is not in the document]
otrosien: Looks like I have to eat my words.

I have the branch checked out locally, and the problem is that at the moment  `T` cannot be assumed to extend Event, so it's impossible to generically reconstruct the event ids of the input batch.

Therefore, any code that allows partially retrying needs to know the concrete type behind T, and the way to fetch the event IDs, and then it's actually not rocket science to iterate through the input events, and check if the event is part of the BatchItemResponse inside the `BatchPublishingException`.

This is why `<T> void onError(List<T> events, Throwable t);` is probably enough.

fahrschein-opentelemetry/src/main/java/org/zalando/fahrschein/opentelemetry/InstrumentedPublishingHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.List;
+import java.util.Locale;
+
+import static io.opentelemetry.api.trace.SpanKind.PRODUCER;
+
+/**
+ * Instruments publishing requests to Nakadi by making use of OpenTelemetry.
+ */
+public class InstrumentedPublishingHandler implements EventPublishingHandler {
```
otrosien, line 17: InstrumentingPublishingHandler?

otrosien (commented):
add documentation in README

MALPI:
👍

RameshMalla:
👍

PR #380, opened 2023-02-22
Add support for zalando platform IAM

RameshMalla:
Our default authorization provider `ZignAccessTokenProvider.java` is outdated and the usage of zign is not recomended anymore by default. 
 
In this PR I have introduced support for [Zalando's platform IAM](https://kubernetes-on-aws.readthedocs.io/en/latest/user-guide/zalando-iam.html) and we could replace `ZignAccessTokenProvider` with this new integration. 

See #381

### DIFF
--- a/fahrschein/build.gradle
+++ b/fahrschein/build.gradle
@@ -12,6 +12,11 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-annotations:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${property('jackson.version')}"
+    compileOnly 'org.projectlombok:lombok:1.18.26'
+    annotationProcessor 'org.projectlombok:lombok:1.18.26'
+
+    testCompileOnly 'org.projectlombok:lombok:1.18.26'
+    testAnnotationProcessor 'org.projectlombok:lombok:1.18.26'
  
     testImplementation ('com.jayway.jsonpath:json-path:2.7.0') {
         exclude group: "org.ow2.asm", module: "asm"
--- a/fahrschein/build.gradle
+++ b/fahrschein/build.gradle
@@ -12,11 +12,6 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-annotations:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${property('jackson.version')}"
-    compileOnly 'org.projectlombok:lombok:1.18.26'
-    annotationProcessor 'org.projectlombok:lombok:1.18.26'
-
-    testCompileOnly 'org.projectlombok:lombok:1.18.26'
-    testAnnotationProcessor 'org.projectlombok:lombok:1.18.26'
  
     testImplementation ('com.jayway.jsonpath:json-path:2.7.0') {
         exclude group: "org.ow2.asm", module: "asm"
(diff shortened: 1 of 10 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/PlatformAccessTokenProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    public PlatformAccessTokenProvider(final String name) {
+        this(DEFAULT_CREDENTIALS_DIRECTORY, name);
+    }
+
+    @Override
+    public String getAccessToken() throws IOException {
+        final Path filePath = Paths.get(this.directory, name + TOKEN_SECRET_SUFFIX);
+        final List<String> lines = Files.readAllLines(filePath, UTF_8);
+        Preconditions.checkArgument(!lines.isEmpty(), "Secret file cannot be empty");
+        return Files.readAllLines(filePath, UTF_8).get(0);
```
RameshMalla, line 55: `ForbiddenApi` is stopping me from using a more straightforward method  `Files.readAllBytes()`

otrosien: you just need to provide a charset :)

fahrschein/build.gradle
```diff
@@ -12,6 +12,11 @@ dependencies {
     implementation "com.fasterxml.jackson.core:jackson-annotations:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${property('jackson.version')}"
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${property('jackson.version')}"
+    compileOnly 'org.projectlombok:lombok:1.18.26'
```
otrosien, line 15: please don't add lombok for this

fahrschein/src/main/java/org/zalando/fahrschein/PlatformAccessTokenProvider.java
```diff
@@ -0,0 +1,57 @@
...
+ *     <dt>{name}-token-type</dt>
+ *     <dd>Contains the token type, e.g. Bearer.</dd>
+ *
+ *     <dt>{name}-token-secret</dt>
+ *     <dd>Contains the actual secret, e.g. a Json Web Token (JWT)</dd>
+ * </dl>
+ *
+ * @see <a href="https://kubernetes-on-aws.readthedocs.io/en/latest/user-guide/zalando-iam.html">Zalando Platform IAM Integration</a>
+ */
+@AllArgsConstructor
```
otrosien, line 29: This is a bad example of lombok usage because it becomes rather implicit to the Lombok usage which one of the two strings of the constructor come first, with high potential for mixing them up. 
And what about using `Path directory`?

fahrschein/src/main/java/org/zalando/fahrschein/PlatformAccessTokenProvider.java
RameshMalla: I thought it was complaining about `Files.readAllBytes(filePath)` 😄 . Thanks 👍

fahrschein/build.gradle
RameshMalla: General idea is to ease development efforts as I can see there were lots of other classes which could have been greatly simplified with Lombok usage. But I can move this discussion as part of another issue. 👍

fahrschein/src/main/java/org/zalando/fahrschein/PlatformAccessTokenProvider.java
RameshMalla: Understood, I was expecting some suggestions over this. The reason why I didn't use `Path` is to reduce the complexity towards our clients, but the constructor mixup could cause serious issues. 👍

fahrschein/build.gradle
RameshMalla: Removed lombok

RameshMalla:
Closing in favor of https://github.com/zalando-nakadi/fahrschein/pull/382

PR #385, opened 2023-03-02
WIP Reverse after and error handler calls

rob-h-w:
So that event publishing handlers do not interact with one another unexpectedly.

## Changes

- Event publishing handlers are called in reverse order onError and afterPublish
- afterPublish is always called regardless of error (finally block) 
- every handler's afterPublish is ensured to be called by adding individual try-catch blocks
- API docs updated

### DIFF
--- a/fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
+++ b/fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
@@ -44,7 +44,6 @@ public class InstrumentedPublishingHandler implements EventPublishingHandler {
     @Override
     public <T> void onError(List<T> events, Throwable t) {
         tracer.activeSpan().setTag(ERROR, true);
-        tracer.activeSpan().finish();
     }
 
     // changes must be applied to both OpenTracing and OpenTelemetry implementations.
--- a/fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
+++ b/fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
@@ -3,6 +3,8 @@ package org.zalando.fahrschein.opentracing;
 import io.opentracing.Span;
 import io.opentracing.Tracer;
 import io.opentracing.tag.Tags;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.zalando.fahrschein.EventPublishingHandler;
 
 import java.util.List;
@@ -12,6 +14,8 @@ import static io.opentracing.tag.Tags.ERROR;
 
 public class InstrumentedPublishingHandler implements EventPublishingHandler {
 
+    private static final Logger LOG = LoggerFactory.getLogger(InstrumentedPublishingHandler.class);
+
     private final Tracer tracer;
 
     public InstrumentedPublishingHandler(Tracer tracer) {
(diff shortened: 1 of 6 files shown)


fahrschein-opentracing/src/main/java/org/zalando/fahrschein/opentracing/InstrumentedPublishingHandler.java
```diff
@@ -44,7 +44,6 @@ public void afterPublish() {
     @Override
     public <T> void onError(List<T> events, Throwable t) {
         tracer.activeSpan().setTag(ERROR, true);
```
otrosien, line 46: why are we not recording the exception in OpenTracing?

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -110,17 +110,23 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
...
-            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onError(events, t));
+            eventPublishingHandlers.descendingIterator().forEachRemaining(handler -> handler.onError(events, t));
             throw t;
         } finally {
             if(response != null) {
                 response.close();
             }
+            eventPublishingHandlers.descendingIterator().forEachRemaining(handler -> {
+                try {
+                    handler.afterPublish();
```
RameshMalla, line 125: Does it mean that `onError` will get executed only when there are failures during `onPublish`? and it looks like any errors while `afterPublish` will be ignored.

otrosien: that is intentional

otrosien: I mean, the actual reason for `onError` is to get called when errors happen in `request.execute()`

otrosien: One could argue that the eventPublishingHandlers.onPublish() should be guarded by their own try-catch blocks  (we once had an incident in Search caused by ungarded span creation and a bug in Lightstep with baggage items handling)

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
```diff
@@ -110,17 +110,23 @@ public <T> void publish(String eventName, List<T> events) throws EventPublishing
...
         Response response = null;
         try {
-            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onPublish(eventName, events));
+            eventPublishingHandlers.forEach(handler -> handler.onPublish(eventName, events));
             response = request.execute();
             LOG.debug("Successfully published [{}] events for [{}]", events.size(), eventName);
-            eventPublishingHandlers.forEach(requestHandler -> requestHandler.afterPublish());
         } catch (Throwable t) {
-            eventPublishingHandlers.forEach(requestHandler -> requestHandler.onError(events, t));
+            eventPublishingHandlers.descendingIterator().forEachRemaining(handler -> handler.onError(events, t));
```
johannesknauft, line 117: If the on-error method creates an exception, we would lose the original error, so at least we should log something before executing the handlers.

johannesknauft: Maybe we should even try/catch the on error execution and just log something if the handler execution failed.

fahrschein/src/main/java/org/zalando/fahrschein/NakadiClient.java
johannesknauft: It all comes down to the point where we need to decide how important the handlers are for the publishing process itself. I have the opinion, that it's the responsibility of the handler to manage their error and our try/catch around this would be just the safety net if they don't do this. Because the client is just responsible for executing the handlers and do the publishing.

otrosien:
👍

MALPI:
👍

johannesknauft:
👍

Issue #384, opened 2023-02-27
Using Subscription Id created with EventTypeSubscription


coolguy1990:
**Is your feature request related to a problem? Please describe.**
With the release of Kubernetes CRD: [EventTypeSubscription](https://nakadi.docs.zalando.net/nakadi-operator/subscription/), we can now create subscription manifests. But we cannot use this Subscription's id directly in the library. As a workaround, we still need to use the same subscription config at code level.

**Describe the solution you'd like**
Ability to stream directly by providing a subscription id.  Instead of the current way of creating a subscription:

```
Subscription subscription = client.subscription(applicationName, eventType)
        .withConsumerGroup(consumerGroup)
        .readFromEnd()
        .withAuthorization(
            authorization()
                .withAdmins(admins)
                .withReaders(readers)
                .build()
        )
        .subscribe()
```

we can have another method probably:

```
Subscription subscription = client.subscriptionId(idFromEventTypeSubscription).subscribe()
```

PR #386, opened 2023-03-05
Using Subscription Id created with EventTypeSubscription

RameshMalla:
See #384

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/domain/Subscription.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/domain/Subscription.java
@@ -26,6 +26,10 @@ public class Subscription {
         this.authorization = authorization;
     }
 
+    public Subscription(String id, String owningApplication, Set<String> eventTypes) {
+        this(id, owningApplication, eventTypes, null, null, null);
+    }
+
 
     public String getId() {
         return id;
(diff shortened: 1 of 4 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/domain/Subscription.java
```diff
@@ -26,6 +28,13 @@ public Subscription(String id, String owningApplication, Set<String> eventTypes,
         this.authorization = authorization;
     }
 
+    //Ignoring this constructor from introspection-based serialization and deserialization functionality,so that It doesn't
+    //interfere with default-all-args constructor.
+    @JsonIgnore
+    public Subscription(String id, String owningApplication, Set<String> eventTypes) {
```
RameshMalla, line 34: Instead of having a new constructor, we can re-use the existing all-args constructor. But given the `public` nature of `Subscription.java` this would give the clients an option to construct their own subscription object without going through `SubscriptionBuilder.java`

Example:

```
Subscription subscription = new Subscription( "subscription-id",
                                                  "application-name",
                                                  Set.of("event-type"));

nakadiClient.stream(subscription)
                .withStreamParameters(new StreamParameters().withBatchLimit(10).withMaxUncommittedEvents(10))
                .listen(TestEvent.class, new TestListener());
```

otrosien:
@RameshMalla no concerns from my end, but please document this new capability in the README.

RameshMalla:
👍

otrosien:
👍

PR #391, opened 2023-03-22
Micrometer Support

### LINKED ISSUE #356: Support micrometer

Spring switched from dropwizard to micrometer. We should consider doing the same, which would actually be simplifying things, as in micrometer instead of injecting a MetricRegistry, you can just register a metric by name whenever you need it, in the same fashion as a logger in SLF4J. 

Consider deprecating and removing fahrschein-dropwizard-metrics, and start using micrometer directly in the code.

**Describe alternatives you've considered**

Alternatively, we can keep the metrics in an extra submodule (e.g. fahrschein-micrometer-metrics), but it takes extra effort for teams to include this dependency, so we won't be instrumented by default. We could as well, bind this submodule as runtime dependency, so it can be pulled in by default. To my understanding it goes a bit against the idea of micrometer, on the other hand teams could actively exclude this dependency if needed for some reason.

**Additional context**

* https://micrometer.io/docs
* https://dzone.com/articles/spring-boot-2-migrating-from-dropwizard-to-micrometer

otrosien:
Adds support for micrometer. 
Closes #356

## Changes
- Extracts the MicrometerMetricsCollector from the spring-boot-starter (impl. code from this branch but default meter prefix + naming based on the spring-boot code)
- Moves MultiplexingMetricsCollector to fahrschein to make it reusable across implementations
- Moves dropwizard implementation from the generic "metrics" package to "metrics.dropwizard" (in java modules, shared packages across jar files would be a violation)
- Creates wrappers for gauge/timer and counter metrics to be used in combination, instead of each implementation individually, and making them package-protected
- Creates Micrometer implementation
- Adds unit tests for both Dropwizard and Micrometer
- Adds documentation (WIP)
- Integrates metrics collector code in example and end-to-end tests
- Drops the getClass() prefix in Dropwizard's LastActivityMetricsCollector for consistency
- Fixes a spelling mistake in the key for `lastErrorHappened`

## TODOs

- [x] Write unit tests for both micrometer and dropwizard.


fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/MicrometerMetricsCollector.java
```diff
@@ -0,0 +1,50 @@
...
+
+import io.micrometer.core.instrument.MeterRegistry;
+import org.zalando.fahrschein.MetricsCollector;
+import org.zalando.fahrschein.MultiplexingMetricsCollector;
+
+import java.util.Arrays;
+
+public class MicrometerMetricsCollector implements MetricsCollector {
+
+    public static final String DEFAULT_PREFIX = "org.zalando.fahrschein.";
```
otrosien, line 11: TODO: Check if a metric name builder exists like in Dropwizard

fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/MicrometerMetricsCollector.java
```diff
@@ -0,0 +1,78 @@
...
+    public void markReconnection() {
+        this.markReconnection();
+    }
+
+    @Override
+    public void markMessageSuccessfullyProcessed() {
+        this.markMessageSuccessfullyProcessed();
+    }
+
+    static String name(String name, String... names) {
```
otrosien, line 58: copied from dropwizard :)

fahrschein-metrics-dropwizard/src/test/java/org/zalando/fahrschein/metrics/dropwizard/DropwizardMetricsCollectorTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Test
+    public void shouldCollectMetrics() {
+        String namespace = "test";
+        MetricRegistry metricRegistry = new MetricRegistry();
+        DropwizardMetricsCollector c = new DropwizardMetricsCollector(metricRegistry, namespace);
+        assertEquals(5, metricRegistry.getMeters().entrySet().size());
+        assertEquals(5, metricRegistry.getGauges().entrySet().size());
+
+        c.markErrorWhileConsuming();
```
johannesknauft, line 22: It would also be cool to test markEventsReceived, as this has a parameter.

fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/MicrometerMetricsCollector.java
```diff
@@ -0,0 +1,78 @@
...
+
+import io.micrometer.core.instrument.MeterRegistry;
+import org.zalando.fahrschein.MetricsCollector;
+import org.zalando.fahrschein.MultiplexingMetricsCollector;
+
+import java.util.Arrays;
+
+public class MicrometerMetricsCollector implements MetricsCollector {
+
+    public static final String DEFAULT_PREFIX = name("fahrschein","listener");
```
johannesknauft, line 11: Is it intended that the prefix is different here?

fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/LastActivityMetricsCollector.java
```diff
@@ -0,0 +1,65 @@
...
+    private long lastMessageSuccessfullyProcessed = 0;
+    private long lastEventReceived = 0;
+    private long lastErrorHappened = 0;
+    private long lastReconnect = 0;
+
+    final Clock clock;
+
+    LastActivityMetricsCollector(final MeterRegistry metricRegistry) {
+        this(metricRegistry, MicrometerMetricsCollector.DEFAULT_PREFIX);
+    }
```
RameshMalla, line 24: With this class being package private, do we need this constructor overload ?

fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/CountingMetricsCollector.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final Counter messagesReceivedMeter;
+    private final Counter eventsReceivedMeter;
+    private final Counter errorsWhileConsumingMeter;
+    private final Counter reconnectionsMeter;
+    private final Counter messagesSuccessfullyProcessedMeter;
+
+    public CountingMetricsCollector(final MeterRegistry metricRegistry) {
+        this(metricRegistry, MicrometerMetricsCollector.DEFAULT_PREFIX);
+    }
```
RameshMalla, line 19: Same as https://github.com/zalando-nakadi/fahrschein/pull/391/files#r1302291218

fahrschein-metrics-micrometer/src/main/java/org/zalando/fahrschein/metrics/micrometer/MicrometerMetricsCollector.java
otrosien: intentional, as the existing micrometer integration into the spring-boot-starter already uses this prefix.

johannesknauft:
:+1:

otrosien:
:+1:

PR #394, opened 2023-04-10
Suppress CVE-2023-20860

otrosien:
Suppress CVE-2023-20860. It doesn't apply to Fahrschein, which is using spring-web as client library.


otrosien:
👍

MALPI:
👍

mesut:
👍

PR #407, opened 2023-07-17
Upgrade Spring dependencies and jdk target

MALPI:
Upgrade to Spring6 and Spring Boot 3.

Thus, we decided to upgrade the JDK target to java17 for the whole project, which justifies a 2.0.0 release.

## Changes

- Bumps baseline JDK to 17
- Sets default dependency for spring framework to 6
- Sets default dependency for spring-boot to 3
- Bumps the version to 2.0.0-RC1

## TODO

- [ ] Mention JDK17 requirement in README
- [ ] Mention Spring version dependencies in README

### DIFF
--- a/fahrschein-http-spring/build.gradle
+++ b/fahrschein-http-spring/build.gradle
@@ -14,7 +14,3 @@ dependencies {
 }
 
 publishing.publications.maven.pom.description = 'Fahrschein HTTP Client using Spring Adapter'
-
-tasks.withType(JavaCompile).configureEach {
-    options.release = 17
-}
--- a/fahrschein-http-spring/build.gradle
+++ b/fahrschein-http-spring/build.gradle
@@ -14,3 +14,4 @@ dependencies {
 }
 
 publishing.publications.maven.pom.description = 'Fahrschein HTTP Client using Spring Adapter'
+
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -16,4 +16,4 @@ opentelemetry.version=1.20.1
 opentracing.version=0.33.0
 
 ## target JDK release version
-jdk.version=8
+jdk.version=17
```
otrosien, line 19: is it required to bump the baseline for the full project or can it be just for the spring-boot-starter (which is already done)?

MALPI: Tried it now by only bumping the spring related modules. 

But looks like we are depending on it as well in the test module, which therefore would require a bump to 17 as well.

fahrschein-http-spring/build.gradle
```diff
@@ -14,3 +14,7 @@ dependencies {
 }
 
 publishing.publications.maven.pom.description = 'Fahrschein HTTP Client using Spring Adapter'
+
+tasks.withType(JavaCompile).configureEach {
```
otrosien, line 18: the fun fact here: the spring-boot-starter is hard-coded against the apache http client, so this one is actually irrelevant. (and even if not, the spring-boot-starter would be able to use it despite being built with java8)

gradle.properties
```diff
@@ -1,12 +1,12 @@
 ## project version
-project.version=1.0.0
+project.version=2.0.0-RC-1
```
otrosien, line 2: RC1

johannesknauft:
:+1:

MALPI:
👍

PR #412, opened 2023-08-02
Remove guava from spring-boot-starter

otrosien:
Trying to reduce the amount of dependencies. The usage of guava in the spring-boot-starter was minimal, and can be replaced by a few lines of static code.

## Changes

- use the already copied "checkArgument" from the core fahrschein module
- replace list functions with ones from Java17
- re-implement camel case and hyphen case conversions with custom implementations (+ tests)
- remove dependency from build.gradle


johannesknauft:
:+1:

otrosien:
👍

PR #414, opened 2023-08-14
Spring-boot-starter: Make http request factory dynamic

otrosien:
## Changes

- drop most of the http client config properties (I think we were overdoing things here. If more control regarding the http client settings is needed, people can still provide their own spring beans.)
- prefer JDK11 request factory if on classpath, otherwise fallback to using the Spring request factory with its own implementation discovery. (`ClientHttpRequestFactories` mentions this priority: HttpComponentsClientHttpRequestFactory,
OkHttp3ClientHttpRequestFactory, SimpleClientHttpRequestFactory)
- Bumps up minimum compatibility (spring boot and spring framework) because we're using `ClientHttpRequestFactories` helpers that got added in spring boot 3.0

This is a breaking change, but IMO a good one, as it adds more control to the spring-boot projects without adding any unwanted dependencies like apache httpclient by default.


otrosien:
TBD: Out-of-the-box experience?? If I upgrade to the latest I will get the spring wrapper using their simple http client implementation (without warning)

mesut:
👍

otrosien:
👍

Issue #417, opened 2023-09-26
Metadata is missing the proper equals and hashcode implementations


fixxer:
**Is your feature request related to a problem? Please describe.**
Effectively immutable class Metadata is missing `equals` and `hashCode`. This causes inconveniences in user code when we implement Event interface, especially in tests when we want to assert two events equality.

```java
    @Test
    public void shouldHaveProperIdentity() {
        final String eventType = "TEST-EVENT-TYPE";
        final String eid = "a3e25946-5ae9-3964-91fa-26ecb7588d67";
        final String partition = "partition1";
        final String version = "v1";
        final String publishedBy = "unauthenticated";
        final OffsetDateTime occurredAt = OffsetDateTime.now(ZoneOffset.UTC);
        final OffsetDateTime receivedAt = OffsetDateTime.now(ZoneOffset.UTC);
        final String flowId = UUID.randomUUID().toString();

        final Metadata metadata1 = new Metadata(eventType, eid, occurredAt, partition, version, publishedBy, receivedAt, flowId, Collections.emptyMap());
        final Metadata metadata2 = new Metadata(eventType, eid, occurredAt, partition, version, publishedBy, receivedAt, flowId, Collections.emptyMap());

        assertEquals(metadata1, metadata2);
    }
```

**Describe the solution you'd like**
Add a proper implementation of the equals and hashcode methods to the Metadata class.

**Describe alternatives you've considered**
Each user implementation of Event may have a custom equals and hashcode implementations checking the Metadata internals. This approach is cumbersome and may be not consistent across the user code.

mesut:
Thank you for your contribution @fixxer

PR #418, opened 2023-09-26
#417 - Implement equals and hashcode for Metadata.

fixxer:
This PR is adding the equals and hashcode methods implementations. It's a fix to the #417 issue.

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
@@ -8,6 +8,7 @@ import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 
 @Immutable
 public final class Metadata {
@@ -81,6 +82,19 @@ public final class Metadata {
 
     public String getPublishedBy() {
         return publishedBy;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Metadata metadata = (Metadata) o;
+        return Objects.equals(eventType, metadata.eventType) && Objects.equals(eid, metadata.eid) && Objects.equals(partition, metadata.partition) && Objects.equals(version, metadata.version) && Objects.equals(publishedBy, metadata.publishedBy) && Objects.equals(occurredAt, metadata.occurredAt) && Objects.equals(receivedAt, metadata.receivedAt) && Objects.equals(flowId, metadata.flowId) && Objects.equals(spanCtx, metadata.spanCtx);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(eventType, eid, partition, version, publishedBy, occurredAt, receivedAt, flowId, spanCtx);
     }
 
     @Override
--- a/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
@@ -108,6 +108,7 @@ public final class Metadata {
                 ", publishedBy='" + publishedBy + '\'' +
                 ", receivedAt=" + receivedAt +
                 ", flowId='" + flowId + '\'' +
+                ", spanCtx=" + spanCtx.toString() +
                 '}';
     }
 }
--- a/fahrschein/src/test/java/org/zalando/fahrschein/domain/MetadataTest.java
+++ b/fahrschein/src/test/java/org/zalando/fahrschein/domain/MetadataTest.java
@@ -7,6 +7,7 @@ import java.time.ZoneOffset;
 import java.util.Collections;
 import java.util.UUID;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class MetadataTest {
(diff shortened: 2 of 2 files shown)


fahrschein/src/main/java/org/zalando/fahrschein/domain/Metadata.java
```diff
@@ -83,6 +84,19 @@ public String getPublishedBy() {
         return publishedBy;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Metadata metadata = (Metadata) o;
+        return Objects.equals(eventType, metadata.eventType) && Objects.equals(eid, metadata.eid) && Objects.equals(partition, metadata.partition) && Objects.equals(version, metadata.version) && Objects.equals(publishedBy, metadata.publishedBy) && Objects.equals(occurredAt, metadata.occurredAt) && Objects.equals(receivedAt, metadata.receivedAt) && Objects.equals(flowId, metadata.flowId) && Objects.equals(spanCtx, metadata.spanCtx);
```
otrosien, line 92: thanks, I just checked - can you add spanCtx to toString as well?

fixxer: Sure, have just done it.

otrosien: hm.. re-reading the PR, do we have a simpler way to test equality? Shouldn't it be enough to use eid (or to be overly-correct eid & eventType & partition)?

otrosien: oh, learned something:

> A useful technique to detect and handle duplicate events on consumer side is to be idempotent and to check eid field of event metadata. Note: eid checking is not possible using the “undefined” category, as it’s only supplied in the “business” and “data” categories.

https://nakadi.docs.zalando.net/nakadi-event-broker/subscriptions/#client-rebalancing

otrosien: on the other hand "undefined" category event types don't even have metadata, so can safely be ignored.

fixxer: Are you suggesting to limit the `equals` implementation only to the `eid` field?

otrosien: yes, I mean from a business perspective two events are the same if their event-id is the same - to reduce (theoretical) collisions we can of course add a bit more, so I would say "eid & eventType & partition & version" - occuredAt, receivedAt, flowId etc should be immutable anyway.

mesut: I think it's a good idea to include all attributes in the hashCode and equals methods. This way, the library can see two Metadata objects as the same when all their attributes match, even if they have different business aspects. It should be up to users to decide if two events are equal or not based on their needs (there might be cases where users want to include certain attributes of events in additional metadata attributes ). They can do this by customizing the hashCode and equals methods in their own event objects.

otrosien: Let's not forget we're still talking about Nakadi. Let's assume this piece of code is used in a hot-path of the application, storing events in a hashmap. Why should we consider other attributes than the obvious ones for identity, wasting CPU cycles for no reason? event-id is a uuid (not sure of which uuid-type), but it's already unique besides some theoretical hash collisions.

otrosien: Regarding event id: This is provided by the producer, therefore you may only draw conclusions within one and the same event type. OTOH this is exactly what the producer considered as "identity" within this stream, therefore my statement holds: identity should be event-type + event-id. (version and partition are actually irrelevant here as well)

To my understanding @fixxer's usecase is equals in a test case, so as long as the test is creating metadata with  eventids following the correct semantics, it should be fine right?

mesut: I appreciate your perspective on this matter. Nakadi doesn't provide a guarantee of uniqueness for event IDs, and this aspect largely depends on the contract in place between producers and consumers. For example, if a producer publishes a batch of events with identical event IDs, Fahrschein should exercise caution in assuming that their metadata and, consequently, the events themselves are identical. This is particularly relevant because events could belong to different partitions and possess varying attributes based on the partition they are associated with. To ensure accuracy, Fahrschein should only consider two metadata objects as equal when all of their attributes match. It's essential that the determination of event equality aligns with the contract established between the producer and consumer, and Fahrschein should refrain from making any unwarranted assumptions regarding event equality.@fixxer concern appears to revolve around maintaining consistent metadata equality when comparing two events, and that's certainly a valid consideration.

Dehora nakadi-java also follows the same approach: https://github.com/dehora/nakadi-java/blob/f3a18c12ba5348de5c66a9ffad3ddb9141296ca3/nakadi-java-client/src/main/java/nakadi/EventMetadata.java#L254

otrosien: hm I get a feeling we're bikeshedding around this topic.  I am willing to "disagree and commit" in this matter.

mesut (approved):

otrosien:
:+1:

fixxer:
I guess there is a hiccup in running the pipeline. Are there any means of properly retriggering it?

otrosien:
will close and reopen this from our main repository.

PR #420, opened 2023-10-04
Remove kotlin dependency & discover jackson modules via classpath introspection

### LINKED ISSUE #419: Remove kotlin dependency from spring-boot-starter
The default jackson object mapper should be instantiated via sth like `Jackson2ObjectMapperBuilder.json().findModulesViaServiceLoader(true).build()` so that the actual context drives it, and we detect scala/kotlin or other runtime environments properly.

otrosien:
The ObjectMapper modules used for Fahrschein can be dicovered from the classpath. This avoids bloating fahrschein dependencies with potential use-cases like kotlin.
 
Closes #419


mesut:
👍

otrosien:
👍

Issue #415, opened 2023-09-08
Nakadi clients resilience to partial outage and partial success


adyach:
**Is your feature request related to a problem? Please describe.**

Nakadi publishing API accepts events in batches. It can fail to publish some events from the batch to underlying storage (Apache Kafka). In that case Nakadi publishing API will return error that batch was partially successful.
It can create problems the following problems, depending on how the Nakadi client and the publishing application deals with this partial success response:
- increase in traffic on Nakadi publishing API due to Nakadi clients retrying the whole batch over and over
- the application retries identical batches which prevents application from progressing

**Describe the solution you'd like**

- Nakadi client should contain a note to developers that publishing can experience partial success. This should be in the client documentation and ideally also within the self contained code documentation, raising awareness for the users, e.g. via docstrings.

- An optional retry method on batch level can be provided for the whole batch, but the default strategy must contain a backoff solution in case of continued errors to publish to Nakadi.

- An optional retry method can be provided that only re-publishes unsuccessful events to Nakadi. This retry must also support a backoff strategy by default.

- Clients must expose the result of a publishing request in a way that developers can understand that there is the possibility of a partial success for batch publishing.

ePaul:
**Current situation**

In case of a partial success (or also in cases like validation errors, which are complete failures), Fahrschein will throw an EventPublishingException with the BatchItemResponses (as returned from Nakadi) for the failed items in the `responses` property.

These objects have the `eid` of the failed event, a `publishingStatus` (failed/aborted/submitted (but these are filtered out)), the `step` where it failed and a detail string.

If the application sets the eids itself (i.e. doesn't let Nakadi do it) and keeps track of them, this allows it to resend only the failed items later. (This is what [nakadi-producer is doing](https://github.com/zalando-nakadi/nakadi-producer-spring-boot-starter/issues/181#issuecomment-1715478983).)
It also allows differentiating between validation errors (which likely don't need to be retried, as they are unlikely to succeed the next time, unless the event type definition is changed) and publishing errors (which should be retried, possibly with some back-off).

But it needs some work on the application side.

**Possible improvements**
_(My personal ideas, I'm not a Fahrschein maintainer.)_

Instead of the plain response from Nakadi, it could help if the actual event object which was submitted is also included in this exception. This makes it easier for an application to retry just these. (This would require Fahrschein to invent eids where the application doesn't provide them, though.)

[...]

PR #424, opened 2023-10-09
Allow clients to correctly handle partial publishing failures

otrosien:
When publishing batches, Nakadi can return partial success. So far, we throw an `EventPublishingException`, independently of a validation vs. persistence error, and we  only include the failed BatchItemResponse entries. This means without event-id correlation locally it is impossible for teams to figure out which items from the batch need retrying. It also not obvious from the error, if the batch can be retried or not.

## Changes

- Removes the filtering for only failed events when crafting the EventPublishingException
- Introduces subclasses for validation vs. persistence errors: `EventPersistenceException` for persistence failures which can be retried, `EventValidationException` for validation errors which need fixing on the client side and should not be retried.
- Adapts the documentation in the README 

## Rationale 

This change allows clients to rely on the element ordering in the thrown exception when considering retrying only the failed events, and properly distinguishing between validation and persistence failures.

Next step towards finishing  #415.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,17 +195,18 @@ Fahrschein does not have sophisticated mechanisms for retry handling when publis
 to handle exceptions when publishing is to create a retry-wrapper around the `NakadiClient.publish` method.
 
 In case of a partial success or also in cases like validation errors, which are complete failures, Fahrschein
-will throw an `EventPublishingException` with the `BatchItemResponse`s included, as returned from Nakadi in the same order as your input.
+will throw an `EventPublishingException`. 
 
-These objects have the eid of the failed event, a `publishingStatus` (failed/aborted/submitted), the step where it failed and a detail string. If your application keeps track of eids, this allows it to resend only the failed items later. Otherwise, you can correlate the failed events based on their position in the sequence.
+You should first call `isRetryable()` on the exception, to identify if it would be allowed to retry publishing the batch.
+
+The exception also has the `BatchItemResponse`s included, as returned from Nakadi in the same order as your input. They also contain the event-ids of the failed events, a `publishingStatus` (failed/aborted/submitted), the step where they failed and a detail string. If your application keeps track of event-ids, this allows it to resend only the failed items later. Otherwise, you can correlate the failed events based on their position in the sequence.
 
 It also allows differentiating between validation errors, which likely don't need to be retried, as they are
 unlikely to succeed the next time, unless the event type definition is changed, and publishing errors
 which should be retried with some back-off.
 
 Recommendation: Implement a retry-with-backoff handler for `EventPublishingException`s, which, depending on
-your ordering consistency requirements, either retries the full batch, or retries the failed events based
-on the event-ids.
+your ordering consistency requirements, either retries the full batch, or retries the failed events. But, ensure to not retry events that failed in the `validating` step.
 
 ## Stopping and resuming streams
 
--- a/README.md
+++ b/README.md
@@ -195,11 +195,9 @@ Fahrschein does not have sophisticated mechanisms for retry handling when publis
 to handle exceptions when publishing is to create a retry-wrapper around the `NakadiClient.publish` method.
 
 In case of a partial success or also in cases like validation errors, which are complete failures, Fahrschein
-will throw an `EventPublishingException` with the `BatchItemResponse`s (as returned from Nakadi) for the failed
- items in the responses property.
+will throw an `EventPublishingException` with the `BatchItemResponse`s included, as returned from Nakadi in the same order as your input.
 
-These objects have the eid of the failed event, a `publishingStatus` (failed/aborted/submitted - but successful itemes are
-filtered out), the step where it failed and a detail string. If the application keeps track of eids, this allows it to resend only the failed items later.
+These objects have the eid of the failed event, a `publishingStatus` (failed/aborted/submitted), the step where it failed and a detail string. If your application keeps track of eids, this allows it to resend only the failed items later. Otherwise, you can correlate the failed events based on their position in the sequence.
 
 It also allows differentiating between validation errors, which likely don't need to be retried, as they are
 unlikely to succeed the next time, unless the event type definition is changed, and publishing errors
--- a/README.md
+++ b/README.md
@@ -194,19 +194,13 @@ Fahrschein supports different exponential backoff strategies when streaming even
 Fahrschein does not have sophisticated mechanisms for retry handling when publishing to Nakadi yet. The recommended way
 to handle exceptions when publishing is to create a retry-wrapper around the `NakadiClient.publish` method.
 
-In case of a partial success or also in cases like validation errors, which are complete failures, Fahrschein
-will throw an `EventPublishingException`. 
-
-You should first call `isRetryable()` on the exception, to identify if it would be allowed to retry publishing the batch.
+In case of a partial success, Fahrschein will throw an `EventPersistenceException`, which is allowed to be retried.
+In case of validation errors - which are complete failures, and should not be retried - it will throw an `EventValidationException`.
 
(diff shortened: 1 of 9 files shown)


otrosien:
@adyach please take a look

fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
```diff
@@ -113,15 +113,15 @@ private void handleAuthError(final JsonNode rootNode, final int statusCode) thro
 
     private void handleBatchItemResponse(JsonNode rootNode) throws IOException {
```
adyach, line 114: is it called only when 207 happens or 422 as well ?

fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
```diff
@@ -113,15 +113,15 @@ private void handleAuthError(final JsonNode rootNode, final int statusCode) thro
...
         final BatchItemResponse[] responses = objectMapper.treeToValue(rootNode, BatchItemResponse[].class);
-        final List<BatchItemResponse> failed = new ArrayList<>(responses.length);
+        boolean partiallyFailed = false;
         for (BatchItemResponse batchItemResponse : responses) {
-            if (batchItemResponse.getPublishingStatus() != BatchItemResponse.PublishingStatus.SUBMITTED) {
-                failed.add(batchItemResponse);
+            if(batchItemResponse.getPublishingStatus() != BatchItemResponse.PublishingStatus.SUBMITTED) {
```
adyach, line 118: lets make it explicit

```suggestion
            if(batchItemResponse.getPublishingStatus() == BatchItemResponse.PublishingStatus.FAILED ||
                batchItemResponse.getPublishingStatus() == BatchItemResponse.PublishingStatus.ABORTED) {
```

fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
```diff
@@ -113,15 +113,15 @@ private void handleAuthError(final JsonNode rootNode, final int statusCode) thro
...
+               partiallyFailed = true;
```
adyach, line 119: why don't you want to throw exception right from here?

adyach (commented):
this changes the current behaviour from having only failed events in the batch response to having all the events in batch response, I do not know if it is documented behaviour.

fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java
otrosien: Both. ProblemHandlingRequest line 38: `if (statusCode == 207 || statusCode >= 400) {`

adyach: ideally it would be different type of exception, but it creates a breaking change in the api as I understand. I would say just keep it like this for now (do not need to extend api with additional flag if users will start to use it will be more work to get rid of it) because it keeps the behaviour the same in this part. it is probably good idea to change this in the future.

otrosien: we're in 2.0.0-RC phase, it's ok to introduce breaking changes now.

adyach: in that case I would go with different exceptions e.g. `EventPersistenceException` and `EventValidationException` because this is the way how it is done at the moment in the library (the way of exceptions). It can be easily added to the documentation of the method.

fahrschein/src/test/java/org/zalando/fahrschein/NakadiClientTest.java
```diff
@@ -302,11 +303,11 @@ public void shouldPublishEventsWithRequestHandler() throws IOException {
 
 
     @Test
-    public void shouldPublishEventsWithRequestHandlerAndFail() throws IOException {
+    public void shouldPublishEventsWithRequestHandlerAndFailWithUnprocessableEntity() throws IOException {
         server.expectRequestTo("http://example.com/event-types/foobar/events", "POST")
                 .andExpectJsonPath("$[0].id", equalTo("1"))
                 .andExpectJsonPath("$[1].id", equalTo("2"))
-                .andRespondWith(207, ContentType.APPLICATION_JSON, "[{\"eid\":\"event-one\",\"publishing_status\":\"failed\",\"step\":\"validating\",\"detail\":\"baz\"}]")
+                .andRespondWith(422, ContentType.APPLICATION_JSON, "[{\"eid\":\"event-one\",\"publishing_status\":\"aborted\",\"step\":\"validating\",\"detail\":\"baz\"}]")
```
adyach, line 310:
```suggestion
                .andRespondWith(422, ContentType.APPLICATION_JSON, "[{\"eid\":\"event-one\",\"publishing_status\":\"failed\",\"step\":\"validating\",\"detail\":\"baz\"}]")
```

only the following events are `aborted`

fahrschein/src/test/java/org/zalando/fahrschein/NakadiClientTest.java
```diff
@@ -319,12 +320,44 @@ public void shouldPublishEventsWithRequestHandlerAndFail() throws IOException {
...
+        verify(eventPublishingHandler, Mockito.atMostOnce()).onError(eq(someEvents), eq(expectedException));
+    }
+
+    @Test
+    public void shouldPublishEventsWithRequestHandlerAndFailPartially() throws IOException {
+        server.expectRequestTo("http://example.com/event-types/foobar/events", "POST")
+                .andExpectJsonPath("$[0].id", equalTo("1"))
+                .andExpectJsonPath("$[1].id", equalTo("2"))
+                .andRespondWith(207, ContentType.APPLICATION_JSON,
+                        "[{\"eid\":\"event-one\",\"publishing_status\":\"failed\",\"step\":\"validating\",\"detail\":\"baz\"},"
```
adyach, line 339:
```suggestion
                        "[{\"eid\":\"event-one\",\"publishing_status\":\"failed\",\"step\":\"publishing\",\"detail\":\"baz\"},"
```

fahrschein/src/test/java/org/zalando/fahrschein/NakadiClientTest.java
```diff
@@ -319,12 +320,44 @@ public void shouldPublishEventsWithRequestHandlerAndFail() throws IOException {
...
+                 + "{\"eid\":\"event-two\",\"publishing_status\":\"submitted\",\"step\":\"none\",\"detail\":\"\"}]")
```
adyach, line 340:
```suggestion
                 + "{\"eid\":\"event-two\",\"publishing_status\":\"submitted\",\"step\":\"publishing\",\"detail\":\"\"}]")
```

README.md
```diff
@@ -194,20 +194,13 @@ Fahrschein supports different exponential backoff strategies when streaming even
 Fahrschein does not have sophisticated mechanisms for retry handling when publishing to Nakadi yet. The recommended way
 to handle exceptions when publishing is to create a retry-wrapper around the `NakadiClient.publish` method.
 
-In case of a partial success or also in cases like validation errors, which are complete failures, Fahrschein
-will throw an `EventPublishingException` with the `BatchItemResponse`s (as returned from Nakadi) for the failed
- items in the responses property.
+In case of a partial success, Fahrschein will throw an `EventPersistenceException`, which is allowed to be retried.
```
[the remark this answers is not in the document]
otrosien: hm.. naming :) I would actually keep it like this - it's understandable and I don't think anyone would think this only refers to a single Event. I also intentionally kept the base exception type `EventPublishingException` to not break existing code.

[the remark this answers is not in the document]
adyach: 👍

otrosien:
@adyach applied suggestions.

otrosien:
👍

adyach:
👍

Issue #219, opened 2018-07-13
Retry on not published events in case of status 207


fbrns:
If event publishing to nakadi returns status 207 at least one event was not published. In this case, retry to send not published events and postpone throwing an exception.

https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L115

jhorstmann:
Would you expect the fahrschein library to do this automatically, or do you mean the todo at https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L124 which would allow the application code to more easily resend failed events?

I don't remember in which situations nakadi would respond with a 207, but I could also imagine that an immediate retry would not work, as nakadi could need some time to recover.

piyushmor:
> Would you expect the fahrschein library to do this automatically, or do you mean the todo at https://github.com/zalando-nakadi/fahrschein/blob/master/fahrschein/src/main/java/org/zalando/fahrschein/ProblemHandlingRequest.java#L124 which would allow the application code to more easily resend failed events?
> 
> I don't remember in which situations nakadi would respond with a 207, but I could also imagine that an immediate retry would not work, as nakadi could need some time to recover.

This can happen during nakadi deployments.

PR #430, opened 2023-10-16
Incorporate retry functionality for event publishing

mesut:
Add retry functionality to event publishing

Fixes the https://github.com/zalando-nakadi/fahrschein/issues/219 issue.

## Publishing Events With Retry

Nakadi publishing API accepts events in batches. It can fail to publish some events from the batch to underlying storage.
In that case Nakadi publishing API will return error that batch was partially successful.
Fahrschein offers retry mechanism complemented by customizable backoff strategies, allowing clients to make a choice between retrying the entire batch or opting for a partial retry as needed. 

For added flexibility, NakadiClient allows configuration of various [backoff strategies](https://github.com/zalando-nakadi/fahrschein#backoff-strategies) for handling event publishing retries.

Fahrschein has the capability to retry the entire batch or only the events within a batch that have failed
(see [available publishing retry strategies below](#available-publishing-retry-strategies)). It will not
retry aborted events, as they failed the event schema validation, and retrying without either changing the event or the 
schema would be futile.

Fahrschein does not enable auto-retry on publishing **by default**. To enable automatic retry on publishing:   

```java

NakadiClient nakadiClient = NakadiClient.builder(NAKADI_URI,new JavaNetRequestFactory(ContentEncoding.GZIP))
        .withPublishingRetryAndBackoffStrategy(PublishingRetryStrategies.FAILED_ONLY, new ExponentialBackoffStrategy()) //default retry configuration
        .build();

nakadiClient.publish("foobar", List.of(
        new SomeEvent("eid1", new Metadata("eid1",OffsetDateTime.now())),
        new SomeEvent("eid2", new Metadata("eid2",OffsetDateTime.now())),
        new SomeEvent("eid3", new Metadata("eid3",OffsetDateTime.now())))
);
```

### Available publishing retry strategies

- `PublishingRetryStrategies.ALL` to retry the entire batch regardless of which events in the batch have failed.
This strategy will lead to high write amplification on Nakadi's side in case of partial outages, and is discouraged.
 
- `PublishingRetryStrategies.FAILED_ONLY` (default) to retry only the events within a batch that have failed. This is  
the recommended default for Nakadi clients, but there is a potential edge case, where within publishing of one batch
Nakadi recovers. If, for example in a batch of three events `[1,2,3]`, all belonging to the same partition, `[1]` fails,
but `[2,3]` succeed, the ordering guarantee for one partition may be potentially be broken, leading to `[2,3,1]` after retry.
For most API clients this is not an issue.
 
- `PublishingRetryStrategies.ALL_FROM_FIRST_FAILURE` to retry the events within a batch, starting with the first failure. 
This strategy trades off write amplification with better ordering guarantees for the aforementioned edge case.
Taking the same example in a batch of three events `[1,2,3]`, all belonging to the same partition, `[1]` fails,
  but `[2,3]` succeed, this retry strategy will retry the full batch of `[1,2,3]`. 

- `PublishingRetryStrategies.NONE` to not retry.

You may also provide your own `PublishingRetryStrategy` implementation.

## TODOs

- [ ] Announce the new default-retry logic for publishing in the release notes


otrosien:
I was discussing with team Aruha that we should have retries enabled by default, wdyt?

mesut:
@otrosien sure, no problem. However, I'd like to understand what the existing `publish` method signature should look like?
How the API should allow clients to opt for full retry or partial retry?  Does it make sense to change the method signature to enable users to pass information about full retry or partial retry? 

Something like this
`public <T> void publish(String eventName, List<T> events, partialRetry) throws EventValidationException, EventPersistenceException, IOException `

Or if we don't want to change the method signature, we can let users to configure that option via NakadiClientBuilder where the users can configure that information. With that approach we would not introduce any breaking changes different from the first option. Something like this=>

`NakadiClientBuilder.withPartialRetry(false).build`

Let me know what do you think?

RameshMalla:
> @otrosien sure, no problem. However, I'd like to understand what the existing `publish` method signature should look like? How the API should allow clients to opt for full retry or partial retry? Does it make sense to change the method signature to enable users to pass information about full retry or partial retry?
> 
> Something like this `public <T> void publish(String eventName, List<T> events, partialRetry) throws EventValidationException, EventPersistenceException, IOException `
> 
> Or if we don't want to change the method signature, we can let users to configure that option via NakadiClientBuilder where the users can configure that information. With that approach we would not introduce any breaking changes different from the first option. Something like this=>
> 
> `NakadiClientBuilder.withPartialRetry(false).build`
> 
> Let me know what do you think?

We can try to enable this feature in `DefaultNakadiPublisher` in spring-boot integration.

otrosien:
A lot of teams are using fahrschein outside of Spring Boot, it should be enabled by default for them as well.

let's add something like: 

`NakadiClientBuilder.withPublishingRetryAndBackoff(PublishingRetryStrategy.NONE|PARTIAL|FULL, BackoffStrategy backoffStrategy)`

and default could be `PublishingRetryStrategy.FULL, new ExponentialBackoffStrategy().withMaxRetries(Integer.MAX_VALUE)`. If no retries are wanted, it can be specified via `.withPublishingRetryAndBackoff(PublishingRetryStrategy.NONE, new NoBackoffStrategy())`

otrosien:
hm.. maybe the question is do consumers re-use the same nakadiclient to publish different events with different settings. In that case your approach is more appropriate.

otrosien:
@ePaul there are event producers which require strict ordering semantics. Imagine publishing two events for the same compaction key, and the first one fails, after that Nakadi recovers and the second event goes through. If you only retry the failed event you will end up with overwriting the second event.

fahrschein/src/main/java/org/zalando/fahrschein/ExponentialBackoffStrategy.java
```diff
@@ -81,4 +81,35 @@ public <T> T call(final int initialExceptionCount, final IOException initialExce
...
+
+        if (count > 0) {
+            sleepForRetries(count);
+        }
+
+        while (true) {
+            try {
+                LOG.warn("Retrying to publish events. Retry count [{}]", count);
+                return callable.call(count, lastRetryException);
+            } catch (EventPersistenceException ex) {
```
mesut, line 103: To retry only in case of EventPersistenceException

otrosien:
@adyach can you take a look and give some more guidance regarding:

- is retry-all on failure a good default choice? (IMO we're optimizing for an edge-case, maybe retry-everything starting from the first failed event in the batch is better? Also: does it make sense to run retry with a batchsize of 1?)
- Instead of using a builder, shall we expose `publishWithRetry` and `publishWithoutRetry` to cater for those two use-cases?

adyach:
> @adyach can you take a look and give some more guidance regarding:
>
> * is retry-all on failure a good default choice? (IMO we're optimizing for an edge-case, maybe retry-everything starting from the first failed event in the batch is better? Also: does it make sense to run retry with a batchsize of 1?)

yes, absolutely. the retry of the whole batch have to be the default option. the retry of failed events can break the order. one concern here about enabling retries by default: how the clients which care about retrying failed events or in some specific cases would understand that starting next version of the library they have to explicitly disable retries or have another strategy implemented. 

> * Instead of using a builder, shall we expose `publishWithRetry` and `publishWithoutRetry` to cater for those two use-cases?

It looks like it is not necessary to change the api contract in that part, that's why I would go with builder / configuration of the policy. that will help users to adopt new version easier (documentation should state how to opt-out from the default behaviour). I did not look into details of the implementation, but I would rather use a single retry policy configuration which state which type retry is being done, e.g. `ExponentialBackoffStrategy` can have a configuration on how it does retry: partial or full. It will make look like common approach to configuration and will not require to expose additional enum variables which makes the contract with the user softer. Builder configuration also looks more fluent and natural (my opinion).

adyach:
> thought when you submit several events to Nakadi in one batch (and they go to the same partition), Nakadi will not publish any events after the first failed one.

@ePaul Nakadi uses non blocking api of Kafka client to send events to brokers. so it will submit all the event from the batch to the Kafka client and will wait for response. Once response for every event is received (or timeout happened) it will aggregate the result. Maybe there is a confusion how validation is done, it is not done for the next events after the event which failed validation.

adyach:
> the retry of failed events can break the order.

apologies, I mixed it up with publishing to available partitions. publishing only failed events won't break the order.

otrosien:
The other objective of this work-package was to avoid write-amplification for clients retrying in case of unavailable partitions (where parts of the batches would continuously go through).

I am still convinced that the best default strategy would be: republish every event of the batch, starting with the first failure individually. In this way a consistent failure of one partition would not lead to repeated partial success, and other issues like too big batch sizes can also be mitigated.

adyach:
> to avoid write-amplification

this is improved by having exponential back off strategy. sending only failed events makes it even better.

ePaul:
> this is improved by having exponential back off strategy.

An exponential back-off strategy only works together with keeping the order if *all submits* (to this event type) are backed off (and kept in order), not just the first failed one.

This sounds like something which can't be implemented on Fahrschein side without changing the whole concept (and the application interface) – you'd need some kind of queue of batches to be submitted, and as currently the submitting thread is waiting until it's done, you'd either get a lot of blocked threads, or need to change to some kind of asynchronous interface. And if the application wants to rely on successful submission, it also will need to change.

ePaul:
> @ePaul Nakadi uses non blocking api of Kafka client to send events to brokers. so it will submit all the event from the batch to the Kafka client and will wait for response. Once response for every event is received (or timeout happened) it will aggregate the result.

So if I submit two events for the same partition in the same batch, it could happen that the first one fails and the second one is published?

If so, then indeed we need to re-publish the whole batch (or, as @otrosien suggests, all events from the first failure), as order will get messed up with any failure.  (Though even then, consumers need to be aware that they can get events out of order, but those events will be corrected by later ones in the right order.)

If not, then re-submitting only the failing events should work just fine.

otrosien:
Summarizing the discussions in this PR, as well as side-track conversations via chat:

Yes, there is an edge case that while sending a batch, a partition in Kafka can come back, so that the ordering guarantee (within that partition) may be violated. Team Aruha is aware of this, and will look into improving the support for this case.

For the time being, we as client of Nakadi are requested to only retry failed events. We will note down this edge case in our docs, so that streams which require stricter ordering guarantees will need to implement different strategies like retry-everything-from-first-failed event (which should be pluggable as well).

fahrschein/src/main/java/org/zalando/fahrschein/ExponentialBackoffStrategy.java
```diff
@@ -81,4 +81,35 @@ public <T> T call(final int initialExceptionCount, final IOException initialExce
             }
         }
     }
+
+    @Override
+    public <T> T call(final EventPersistenceException lastException, final int retryCount,
```
otrosien, line 86: smell: why do we have to copy-paste the code from above?

otrosien:
TODOs:
- do not auto-enable retry logic
- avoid copy-paste code in ExponentialBackoffStrategy

fahrschein/src/main/java/org/zalando/fahrschein/ExponentialBackoffStrategy.java
```diff
@@ -81,4 +80,35 @@ public <T> T call(final int initialExceptionCount, final IOException initialExce
             }
         }
     }
+
+    @Override
+    public <T> T call(final int initialExceptionCount, final EventPersistenceException initialException,
```
mesut, line 85: I attempted to consolidate the 'call' methods to reduce duplication, but it ended up making the code more complex, which, in my opinion, adversely affects readability. I'm open to suggestions.

```
private <T> T performRetry(final int initialExceptionCount, final IOException initialException,
            final Object callable) throws BackoffException, InterruptedException {
        checkMaxRetries(initialException, initialExceptionCount);
        int count = initialExceptionCount;
        if (count > 0) {
            sleepForRetries(count);
        }
        IOException lastRetryException = initialException;
        while (true) {
            try {
                LOG.trace("Try [{}]", count);
                if (callable instanceof ExceptionAwareCallable<?>) {
                    return ((ExceptionAwareCallable<T>) callable).call(count,
                            (EventPersistenceException) lastRetryException);
                }
                return ((IOCallable<T>) callable).call();
            } catch (EventPersistenceException e) {
                lastRetryException = e;
            } catch (IOException e) {
                lastRetryException = e;
                if (callable instanceof ExceptionAwareCallable<?>) {
                    throw new BackoffException(e, count);
                }
            } finally {
                LOG.warn("Got [{}] on retry [{}]", lastRetryException.getClass()
                        .getSimpleName(), count, lastRetryException);
                count++;
                checkMaxRetries(lastRetryException, count);
                sleepForRetries(count);
            }
        }
    }
```

fahrschein/src/main/java/org/zalando/fahrschein/ExponentialBackoffStrategy.java
mesut: @otrosien I attempted to consolidate the 'call' methods to reduce duplication, but it ended up making the code more complex, which, in my opinion, adversely affects readability. I'm open to suggestions.

```
private <T> T performRetry(final int initialExceptionCount, final IOException initialException,
            final Object callable) throws BackoffException, InterruptedException {
        checkMaxRetries(initialException, initialExceptionCount);
        int count = initialExceptionCount;
        if (count > 0) {
            sleepForRetries(count);
        }
        IOException lastRetryException = initialException;
        while (true) {
            try {
                LOG.trace("Try [{}]", count);
                if (callable instanceof ExceptionAwareCallable<?>) {
                    return ((ExceptionAwareCallable<T>) callable).call(count,
                            (EventPersistenceException) lastRetryException);
                }
                return ((IOCallable<T>) callable).call();
            } catch (EventPersistenceException e) {
                lastRetryException = e;
            } catch (IOException e) {
                lastRetryException = e;
                if (callable instanceof ExceptionAwareCallable<?>) {
                    throw new BackoffException(e, count);
                }
            } finally {
                LOG.warn("Got [{}] on retry [{}]", lastRetryException.getClass()
                        .getSimpleName(), count, lastRetryException);
                count++;
                checkMaxRetries(lastRetryException, count);
                sleepForRetries(count);
            }
        }
    }
```

otrosien: I'll create a follow-up PR for the cleanup. The current PR is already pretty big.

otrosien:
👍

mesut:
👍

PR #432, opened 2023-10-18
StreamParameters refresher (documentation + new parameter support)

### LINKED ISSUE #431: Support stream parameter "batch_timespan"

Nakadi has a stream-parameter called batch_timespan, which is currently unknown to Fahrschein.

> batch_timespan: Useful for batching events based on their received_at timestamp. Nakadi would flush a batch as soon as the difference in time between the first and the last event in the batch exceeds the batch_timespan.

Default is '0', meaning "do not inspect timestamps".

Fahrschein should support this optional parameter, both within or outside of Spring Boot.

otrosien:
## Changes

- Supports stream parameter "batch_timespan" (closes #431)
- updates and add API documentation based on the latest API docs from Nakadi, incl. documentation of defaults
- Removes client-side validation of parameters (duplicates what the server will reject anyway)
- Removes an outdated comment from 2017 about a "silent bug". Nowadays Nakadi rejects a batch_limit of 0. (Verified and tested with Nakadi team)

### DIFF
--- a/fahrschein/src/main/java/org/zalando/fahrschein/StreamParameters.java
+++ b/fahrschein/src/main/java/org/zalando/fahrschein/StreamParameters.java
@@ -8,8 +8,6 @@ import java.util.stream.Collectors;
 
 public class StreamParameters {
 
-    private static final int DEFAULT_BATCH_LIMIT = 1;
-
     @Nullable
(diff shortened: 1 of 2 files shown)


otrosien:
👍

mesut:
👍

PR #444, opened 2023-11-03
Autowire tracing in spring boot

### LINKED ISSUE #443: Spring Boot: Autowire OpenTelemetry integration

Setting up Fahrschein with OpenTelemetry enabled requires manual effort:
- defining your own NakadiClient beans
- attaching the `InstrumentedPublishingHandler` as request handler when building the NakadClient.

If OpenTelemetry is autowired in Spring Boot (we can check for a `Tracer` bean), we should be able to set up Fahrschein accordingly.

**Describe alternatives you've considered**

Adding more detailed information into the README.md.

otrosien:
## Changes

- Registers the `InstrumentedPublishingHandler` as bean, in case a Tracer bean from openTelemetry is available in the spring context
- Adds the InstrumentedPublishingHandler to the factory building the NakadiClient
- Cleans up useless code in `MeterRegistryAwareBeanPostProcessor` (`postProcessAfterInitialization`)
- Adds a test 

Closes  #443

### DIFF
--- a/fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
+++ b/fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
@@ -123,7 +123,6 @@ public class FahrscheinRegistrar implements NakadiClientsRegistrar {
         }
     }
 
-
     private void registerAliasesForNakadiListener() {
         registry.registerAliasesForNakadiListener();
     }


fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/MeterRegistryAwareBeanPostProcessor.java
```diff
@@ -28,8 +28,4 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
         this.beanFactory = beanFactory;
     }
 
-    @Override
-    public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
```
otrosien, line 32: not used.

otrosien:
👍

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
```diff
@@ -65,15 +66,16 @@ private String registerPublisher(PublisherConfig publisherConfig) {
...
+            final String requestFactoryRef = registerRequestFactory(config);
             return genericBeanDefinition(FahrscheinNakadiClientFactory.class)
-                .addConstructorArgValue(consumerConfig)
-                .addConstructorArgReference(registerCursorManager(consumerConfig, requestFactoryRef))
-                .addConstructorArgReference(registerObjectMapper(consumerConfig))
+                .addConstructorArgValue(config)
+                .addConstructorArgReference(registerCursorManager(config, requestFactoryRef))
+                .addConstructorArgReference(registerObjectMapper(config))
                 .addConstructorArgReference(requestFactoryRef)
+                .addConstructorArgValue(registry.getBeansOfType(EventPublishingHandler.class).values().stream().toList())
```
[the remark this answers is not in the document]
RameshMalla: `Picks up all PublishingHandlers and adds them to the NakadiClient`
I just saw the above point in the PR description.

I would suggest to load only the required handlers specified by the client via config file + `InstrumentedPublishingHandler`, instead of loading all handler implementations.   This would keep the library more flexible and also will allow clients to use specific handlers for specific publishers.

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
```diff
@@ -65,15 +66,16 @@ private String registerPublisher(PublisherConfig publisherConfig) {
         });
     }
 
-    private String registerNakadiClient(AbstractConfig consumerConfig, String type) {
-        return registry.registerIfAbsent(consumerConfig.getId() + "-" + type, NakadiClient.class, () -> {
-            log.info(LOG_PREFIX + "NakadiClient ...", consumerConfig.getId());
-            final String requestFactoryRef = registerRequestFactory(consumerConfig);
+    private String registerNakadiClient(AbstractConfig config, String type) {
```
RameshMalla, line 69: The same method is used for registering the Nakadi client by both consumers and publishers, but we only require instrumentation for publishers. Perhaps it's time to consider creating dedicated clients for each role – one for consumers and one for publishers.

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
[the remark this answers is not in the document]
otrosien: IMO that would mean optimizing for some weird edge case, assuming that a) the client implements its own `PublishingHandler`, and b) registers multiple publishers which would require a different set of `PublishingHandlers`.

The current code would "just work" for the common-sense setup, even on custom `PublishingHandler` implementations, which is IMO the assumption for Spring Boot Autostarters.

How would you envision a client registering its own `PublishingHandler` in your approach? They would still have to create their own `NakadiClient` bean, and do all the wiring by hand, right? So, there is technically nothing gained by not pulling all `PublishingHandlers` by default.

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
otrosien: agree, but too big of a breaking change for now.

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/config/Registry.java
```diff
@@ -110,6 +111,15 @@ public static <T> List<T> list(final T... elements) {
         return list;
     }
 
+    public <T> Map<String, T> getBeansOfType(Class<T> clazz) {
+        if (registry instanceof DefaultListableBeanFactory) {
+            DefaultListableBeanFactory factory = (DefaultListableBeanFactory) registry;
+            return factory.getBeansOfType(clazz);
+        } else {
+           LOG.warn("Unable to get beans of type {} from registry of type: {}", clazz.getName(), registry.getClass().getName())
```
otrosien, line 119: @RameshMalla actually, I would consider these errors (more drastically in `registerAliasesForNakadiListener` below: If we cannot pick up NakadiEventListener annotated beans, the whole application setup is broken..)

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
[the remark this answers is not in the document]
RameshMalla: By default, loading the `InstrumentedPublishingHandler` in the presence of a `Tracer` makes sense, since we are the owners of this implementation. It offers the client a hassle-free, commonly-used setup without requiring manual configuration.

However, if we automatically load all other custom handlers without the client's consent, we risk introducing side effects, especially in scenarios involving multiple publishers (though not typical, still possible). Additionally, maintaining the execution order of handlers would become non-deterministic in such cases.

> Clients would still need to create their NakadiClient bean and perform the necessary wiring manually.

To address this, we can provide clients with a provision to specify handler beans in the configuration file, as demonstrated below:

```
fahrschein:
  publishers:
     handlers:
          - handlerOne
          - handlerTwo
```

This allows clients to define the handlers they want to use.

IMHO, the decision to load all request handlers can be deferred for now, allowing us to further refine the use case. Additionally, we need to undertake some refactoring in the existing code to ensure it is free from side effects before introducing this feature.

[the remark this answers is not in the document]
otrosien: hm, ok. Will push an update to the PR.

fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinRegistrar.java
```diff
@@ -121,6 +123,7 @@ private void registerDefaultObjectMapper() {
         }
     }
 
+
```
otrosien, line 126:
```suggestion
```

RameshMalla (approved):
LGTM

RameshMalla:
👍

otrosien:
:+1:

PR #449, opened 2023-12-01
Fix CVE-2023-34053 by upgrading to spring 6.0.14

otrosien:
## Changes

- upgrade spring to 6.0.14 and spring-boot to 3.1.6


otrosien:
👍

Issue #450, opened 2023-12-07
Spring Boot: FahrscheinNakadiConsumer ignores consumer properties when subscription-id is not provided


levabski:
**Describe the bug**
`FahrscheinNakadiConsumer` relies on `consumer-group` and `authorization` properties to establish a subscription in the absence of a provided `subscription-id`. However, the `SubscriptionBuilder` constructs an object without assignment, leading to exceptions being thrown.
https://github.com/zalando-nakadi/fahrschein/blob/46548db81f9d614eeba0d9711d58f07d43f9553e/fahrschein-spring-boot-starter/src/main/java/org/zalando/spring/boot/fahrschein/nakadi/config/FahrscheinNakadiConsumer.java#L141-L151

**To Reproduce**
1. Create application.yaml file with the following configuration:
```
fahrschein:
  defaults:
    nakadi-url: ${NAKADI_URL}
    application-name: ${NAKADI_APPLICATION_NAME}
    consumer-group: ${NAKADI_CONSUMER_GROUP}
    authorization:
      anyReader: false
      readers:
        users:
          - ${NAKADI_USER}
    oauth:
      enabled: true
      access-token-id: ${NAKADI_ACCESS_TOKEN_ID}
  consumers:
    ${NAKADI_CONUSMER_ID}:
      topics:
        - ${NAKADI_EVENT_TYPE_NAME}
```

**Environment:**
 - Eclipse Temurin 21.0.1
 - org.zalando:fahrschein-spring-boot-starter:2.0.0-RC4

PR #451, opened 2023-12-07
Spring Boot: Add Subscription config properties when subscription-by-id is not provided

levabski:
Spring Boot: FahrscheinNakadiConsumer ignores consumer properties when subscription-id is not provided https://github.com/zalando-nakadi/fahrschein/issues/450


otrosien:
👍

johannesknauft:
:+1:

PR #454, opened 2023-12-19
Upgrade dependencies to latest versions

otrosien:
Supersedes #449 (Fix for CVE-2023-34053)

## Changes 

- Upgrade dependencies to latest versions. 
- Fixes the issue in current CI runs, that the taskScheduler bean in our ExampleApplication is invalid, and rejected in latest version of spring boot:

This bean must be a left-over of spring boot 2.x days, in spring boot 3 this is not needed, according to API docs of `SchedulingConfigurer`: 

> SchedulingConfigurer
Optional interface to be implemented by [@Configuration](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html) classes annotated with [@EnableScheduling](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/EnableScheduling.html). Typically used for setting a specific [TaskScheduler](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/TaskScheduler.html) bean to be used when executing scheduled tasks or for registering scheduled tasks in a programmatic fashion as opposed to the declarative approach of using the [@Scheduled](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/annotation/Scheduled.html) annotation. For example, this may be necessary when implementing [Trigger](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/Trigger.html)-based tasks, which are not supported by the @Scheduled annotation.

We don't use Trigger-based tasks.

## Logs

```
BeanDefinitionOverrideException: Invalid bean definition with name 'taskScheduler' defined in class path resource [org/springframework/boot/autoconfigure/task/TaskSchedulingConfigurations$TaskSchedulerConfiguration.class]
```


johannesknauft:
:+1:

PR #461, opened 2024-01-24
CVE-2024-22233: Upgrade to latest spring framework and spring boot.

otrosien:
Fahrschein does not put spring security on the classpath, so technically this is a false positive. It's nevertheless adivsable to upgrade to the latest non-vulnerable version of spring.


otrosien:
👍

mesut:
👍

PR #463, opened 2024-01-26
Skip codacy check from forks

otrosien:
There are security constraints so that we cannot use a secret in github actions from forks. This makes the codacy step fail.

## Change

- Makes the execution of codacy conditional to only PRs from non-forks.

## Alternatives considered

- Allowing exposure of secrets to PRs from forks can only be org-centrally enabled. If we had this enabled, and manual approval of running Github workflows for forks, we would be good. But it's unlikely to happen due to risks of exposing sensitive secrets.


rob-h-w:
:+1:

rameshmalla:
👍

PR #466, opened 2024-02-23
Improve logging information for failed events during parsing.

as-belcastro:
### Context
In case "alien" events pop up in Nakadi event type, it can't be parsed by the client and issues are raised by the EventReader. Information about the partition and offset are not logged out and troubleshooting it is quite difficult. Therefore this change identifies issues during parsing/reading of events and allows the read batch to finish it process/loop so the cursor information can be grabbed and printed out afterwards.


otrosien:
Hey, thanks for the contribution. Is it possible to write a test for this?

as-belcastro:
> Hey, thanks for the contribution. Is it possible to write a test for this?

I've added a test that checks the exception and the log message itself.

mchernyavskaya:
👍

as-belcastro:
👍

otrosien:
👍

johannesknauft:
:+1:

